PR #258, opened 2014-12-19
raise an error if trying to create a monetized attribute with the same name as an existing attribute

george-carlin:


TiuTalk:
Why this exception is necessary? Can't the method just fail silently?

I'm asking because this commit is giving me a lot of problems when booting my application via, lets say, Sidekiq, where it normally load/initialize the application twice.. And when the model will be loaded again, it stops on the `monetize` lines.

semmons99:
Fail silently is never a good thing. If needed, we should account for your situation. Feel free to research and submit a PR.

PR #266, opened 2015-01-03
Fix store_accessor compatibility

### LINKED ISSUE #263: Doesn't work with store_accessor under Rails 4.2
``` ruby
create_table 'services' do |t|
  t.text 'data'
end

class Service < ActiveRecord::Base
  store :data, accessors: [:foo_cents]
  monetize :foo_cents, as: :foo
end

> s = Service.new
> s.foo_cents = 100
=> 100
> s.foo
=> #<Money fractional:100 currency:USD>
> s.foo = 100
ActiveModel::MissingAttributeError: can't write unknown attribute `foo_cents`
    from /home/ari/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/activerecord-4.2.0/lib/active_record/attribute.rb:124:in `with_value_from_database'
...
    from /home/ari/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/activerecord-4.2.0/lib/active_record/attribute_methods/dirty.rb:92:in `write_attribute'
    from /home/ari/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/money-rails-1.2.0/lib/money-rails/active_record/monetizable.rb:174:in `block in monetize'
```

This works fine under Rails 4.1, but not 4.2. If I change monetizable.rb:174 to `public_send("#{subunit_name}=", money.try(:cents))` instead, it seems to work fine.

I can't tell if this is a Rails bug or a money-rails bug.

aripollak:
* Fixes #263 
* In the process of refactoring, I also found a bug in the validator that I fixed - if you set the fractional amount directly, the validator on the money object uses the fractional amount instead of the real money value.

### DIFF
--- a/lib/money-rails/active_model/validator.rb
+++ b/lib/money-rails/active_model/validator.rb
@@ -17,10 +17,11 @@ module MoneyRails
         @raw_value = @record.try(before_type_cast)
 
         # If raw value is nil and changed subunit is nil, then
-        # nil is a assigned value, elsewhere we should treat the
+        # nil is a assigned value, else we should treat the
         # subunit value as the one assigned.
-        if @raw_value.nil? && @record.send(subunit_attr)
-          @raw_value = @record.send(subunit_attr)
+        if @raw_value.nil? && @record.public_send(subunit_attr)
+          subunit_value = @record.public_send(subunit_attr)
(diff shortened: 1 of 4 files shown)


semmons99:
thanks!

PR #269, opened 2015-01-09
Add note about default currency into readme.

Olefine:
We need note about default currency, because in most cases you do next steps
1) Add gem into gemfile
2) Create model or add money migration
3) Migrate db
4) Change defaults in initializer file.

In Our case, change defaults in initializer after migration, changes will not work

We can create rails generator for configuring money in rails.


semmons99:
Thanks!

PR #278, opened 2015-01-26
Using currency_column[:postfix] to automatically determine currency column

antstorm:
Obviously the `:postfix` option for `currency_column` was not used at all, so I decided to implement it.

Basically it will try to find an attribute based on `name` and `currency_column[:postfix]` and if there is one — use it to get currency value.

I also removed `public_send(instance_currency_name).present?` because it is redundant and  `Money::Currency.find` can handle empty strings and `nil`s itself, returning `nil` in those cases.

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -207,14 +207,19 @@ module MoneyRails
             end
 
             define_method "currency_for_#{name}" do
+              instance_currency_name_with_postfix = "#{name}#{MoneyRails::Configuration.currency_column[:postfix]}"
+
               if instance_currency_name.present? &&
                 respond_to?(instance_currency_name) &&
-                public_send(instance_currency_name).present? &&
                 Money::Currency.find(public_send(instance_currency_name))
 
                 Money::Currency.find(public_send(instance_currency_name))
               elsif field_currency_name
                 Money::Currency.find(field_currency_name)
+              elsif respond_to?(instance_currency_name_with_postfix) &&
+                Money::Currency.find(public_send(instance_currency_name_with_postfix))
+
+                Money::Currency.find(public_send(instance_currency_name_with_postfix))
               elsif self.class.respond_to?(:currency)
                 self.class.currency
               else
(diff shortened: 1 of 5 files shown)


aganov:
@antstorm @semmons99 With merge applied now, `:postfix` is only used for reading, should't we apply the same logic when setting the monetized value?


```ruby
# Current behavior
MoneyRails.configure do |config|
  config.default_currency = :gbp
  config.currency_column = {
    # ...
    postfix: "_currency",
  }
end

# t.integer :amount_cents
# t.string :amount_currency
class Transaction < ActiveRecord::Base
  monetize :amount_cents
end

#  :amount_cents is blank, and amount is fetched with default_currency
t = Transaction.new(amount: "5 EUR")
# t.amount_cents => nil
# t.amount => #<Money fractional:500 currency:GBP>
```

https://github.com/RubyMoney/money-rails/blob/master/lib/money-rails/active_record/monetizable.rb#L187-L196

Patch?

```ruby
if MoneyRails::Configuration.currency_column[:postfix].present?
    instance_currency_name_with_postfix = "#{name}#{MoneyRails::Configuration.currency_column[:postfix]}"
end

# Update currency iso value if there is an instance currency attribute
if instance_currency_name.present? && respond_to?("#{instance_currency_name}=") && money_currency
  public_send("#{instance_currency_name}=", money_currency.try(:iso_code))
elsif instance_currency_name_with_postfix.present? && respond_to?("#{instance_currency_name_with_postfix}=") && money_currency
  public_send("#{instance_currency_name_with_postfix}=", money_currency.try(:iso_code))
else
  current_currency = public_send("currency_for_#{name}")
  if money_currency && current_currency != money_currency.id
    raise ReadOnlyCurrencyException.new("Can't change readonly currency '#{current_currency}' to '#{money_currency}' for field '#{name}'") if MoneyRails.raise_error_on_money_parsing
    return nil
  end
end
```

antstorm:
@aganov hmm… that's quite odd… which version are you using?

aganov:
@antstorm money-rails (1.6.0)

PR #280, opened 2015-01-28
Update README.md

mrbongiolo:
Explaining that :no_cents_if_whole defined on config.default_formats is overwritten by config.no_cents_if_whole. And humanized_money will "ignore" it, since config.no_cents_if_whole is nil, so it will set it to true.


alup:
thanks

PR #285, opened 2015-02-13
Fix .monetized_attributes class method

akoltun:
If both superclass and subclass uses monetize macros then .monetized_attributes has to merge recurrently @monetized_attributes from all superclasses.

[This commit](https://github.com/RubyMoney/money-rails/commit/8c2a2d45bb0f96f7b747c60de3dc4dab094835f5) has revealed this bug


semmons99:
nice!

PR #292, opened 2015-02-18
Field currency values can be determined by lambda

KazW:
Hi, I think I could add a few more tests to this PR (I only added one). If anyone has any advice as to where and how to best test this, please comment here.

Thanks to the authors and contributors for making this awesome gem!


semmons99:
looks solid to me!

PR #297, opened 2015-03-02
Simplify validation options

antstorm:
Much simpler passing of validation options.

The only difference is that passing `subunit_numericality: false` won't work, but I'm not sure if this was ever a feature, because there's no docs or tests for this behaviour.

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -90,16 +90,20 @@ module MoneyRails
             if validation_enabled = MoneyRails.include_validations && !options[:disable_validation]
 
               # This is a validation for the subunit
-              validates subunit_name, {
-                :allow_nil => options[:allow_nil],
-                :numericality => options[:subunit_numericality] || true
-              }
+              if subunit_numericality = options.fetch(:subunit_numericality, true)
+                validates subunit_name, {
+                  :allow_nil => options[:allow_nil],
+                  :numericality => subunit_numericality
+                }
+              end
 
               # Allow only Money objects or Numeric values!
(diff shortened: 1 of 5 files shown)


lib/money-rails/active_record/monetizable.rb
```diff
@@ -89,30 +89,16 @@ def monetized_attributes
...
-                true
-              else
-                options[:numericality]
-              end
-
               # This is a validation for the subunit
               validates subunit_name, {
                 :allow_nil => options[:allow_nil],
-                :numericality => subunit_validation_options
+                :numericality => options[:subunit_numericality] || true
```
pftg: what do you think about using `Hash#fetch` then `false` will be worked?
By the way I think we does not have tests for `false` values, so you example passed but should not.

pftg:
@antstorm please add tests for case like: `subunit_numericality: false`

lib/money-rails/active_record/monetizable.rb
semmons99: Agreed. Fetch is our friend here. 



- shane

antstorm:
@pftg for some reason I keep forgetting the handy `fetch` method, thanks for reminding!

I've added tests for skipping validations separately.

semmons99:
Still getting failures. Post a message here if you need help.

antstorm:
@semmons99 yeah, it seems like Rails 3 doesn't care about what you pass to `:numericality` in `validates`. As long as the key is present it will run the validation. Hopefully that will fix it. Seems like the initial version would also fail those new tests, which is good, because now it's fixed.

semmons99:
merged!

PR #299, opened 2015-03-05
Always support t.monetize in migrations

myShoggoth:
In mixed DB environments (in my case, sqlite3 for dev and postgres for production), migrations would work in one but not the other, which is not acceptable.

This change makes t.monetize/add_monetize/remove_monetize always available, and supports the 'money' version in non-postgres environments for backwards compatibility.


semmons99:
:metal:

PR #300, opened 2015-03-14
Refactor monetized_attributes

antstorm:
A bit of refactoring for the `monetized_attributes`. I think this makes it more straightforward and lightens up the heavy `monetize` method a bit.

Also added the test for overriding a monetized attribute in a subclass (which was also possible to do before this change).


semmons99:
awesome!

PR #304, opened 2015-03-22
Include postgres specific code also when adaptor = postgis.

undergroundwebdesigns:
Allows the Posgres specific migrations to be used in Rails apps that use the ActiveRecord postgis adapter.


semmons99:
woot! thanks.

PR #313, opened 2015-04-09
chore(add read only exception class)

zhubert:
We're finding that it would be really nice to be able to rescue the exception which occurs when a currency  that is readonly undergoes an attempted modification.  

For instance, user types "35EUR" into a field which is in dollars, we'd like to catch that exception and make it into a helpful user message.


semmons99:
Nice!

PR #317, opened 2015-04-20
Extended validator to reject string values where a thousand separator...

touchardv:
...appears after the decimal mark.


semmons99:
:metal:

PR #326, opened 2015-05-26
Include attribute in validation messages like Rails does

chewi:
Always having the attribute name anchored to the start of the validation message is quite restrictive so I redefine them in such a way that the attribute name can be placed anywhere. This relies on the `:attribute` key being set, which Rails does but MoneyRails doesn't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## master (next release)
 
- - Respect Money.use_i18n when validating.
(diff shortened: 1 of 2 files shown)


antstorm:
Looks good to me!

semmons99:
update the changelog and I'm okay merging.

chewi:
I'll do that when you've merged the other request so that it applies cleanly.

chewi:
Done.

PR #329, opened 2015-05-30
Update README.md

wykhuh:
Minor grammar change. 

Changed
``` 
If the currency of a field with vary from once model instance to another
```
to
```
 If the currency of a field will vary from one model instance to another
```


antstorm:
Thanks, @wykhuh!

Although I think that `might` or `can` instead of `will` is more appropriate here. Or maybe `varies` instead of `will vary`. Correct me if I'm wrong…

PR #332, opened 2015-06-08
Respect raise_error_on_money_parsing before raising ReadOnlyCurrencyException

chewi:
If the user gives a currency when they're not supposed to then it
makes sense to react as though the value is generally invalid.

I can merge this myself but I'd like some feedback first.


semmons99:
Seems reasonable to me. Any objections @RubyMoney/money-rails-devs?

antstorm:
Looks good to me!

PR #333, opened 2015-06-10
Only do additional validation if value is a number

chewi:
If the value given turns out not to be a number then it doesn't make sense to show both the "must be a valid currency" and "not a number" messages.

There's also some additional clean-up here. Feedback appreciated. :smile:


semmons99:
:metal:

PR #336, opened 2015-06-24
Prevent infinite loop when currency postfix is nil

### LINKED ISSUE #286: Stack level too deep during tests
I'm having a bug very similar to #38, after I add monetize to my model and I run `rspec model.rb` it hangs throwing a stack level too deep and I can see that ruby goes in disk sleep, nothing else opens and my only option is to reboot.

I've identified the problem in my config setting:

```
config.currency_column = { prefix: '',
                              postfix: '',
                              column_name: 'currency',
                              type: :string,
                              present: true,
                              null: false,
                              default: 'USD'
                            }
```

Without this it works, with it it breaks. In production/development it works correctly though: when there is the currency column it uses it, otherwise it uses the default USD.

I have various models, and not all of them have the currency column eg:

- model1:
  - price_1
  - price_2
  - currency
- model2 (the one I'm currently testing):
  - price
- model 3:
  - price
  - currency

I've also created a quick app to test it out and it confirmed that the problem is when that config is added

```
  create_table "posts", force: :cascade do |t|
    t.integer  "price_cents"
    t.string   "title"
    t.datetime "created_at",  null: false
    t.datetime "updated_at",  null: false
  end

class Post < ActiveRecord::Base
  validates :price_cents, numericality: { greater_than: 0.0 }
  validates :title, presence: true

  monetize :price_cents

end

require 'rails_helper'

RSpec.describe Post, type: :model do
  it { is_expected.to validate_numericality_of(:price_cents).is_greater_than(0) }
  it { is_expected.to monetize(:price_cents) }

  it { is_expected.to validate_presence_of :title }
end

```

sslotnick:
Fixes #286 (looping in @antstorm who made the original change)

When the currency column postfix is nil then `instance_currency_name_with_postfix` ends up being the same as the field name. 

Line 211 (old) was causing it to call the field name not the currency, creating an infinite loop between `currency_for_#{name}` and the field accessors.

I'm happy to add tests if someone can provide guidance on how to properly override the currency column name.

Note: This requires the instance currency field to be nil, otherwise line 211 (new) will return it. I'm not sure if it is desired behavior, but the default currency is not populated on instantiation, which allows this to happen.

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -197,7 +197,9 @@ module MoneyRails
             end
 
             define_method "currency_for_#{name}" do
-              instance_currency_name_with_postfix = "#{name}#{MoneyRails::Configuration.currency_column[:postfix]}"
+              if MoneyRails::Configuration.currency_column[:postfix].present?
+                instance_currency_name_with_postfix = "#{name}#{MoneyRails::Configuration.currency_column[:postfix]}"
+              end
 
               if instance_currency_name.present? && respond_to?(instance_currency_name) &&
                   Money::Currency.find(public_send(instance_currency_name))
@@ -207,8 +209,9 @@ module MoneyRails
                 Money::Currency.find(field_currency_name.call(self))
               elsif field_currency_name
                 Money::Currency.find(field_currency_name)
-              elsif respond_to?(instance_currency_name_with_postfix) &&
-                  Money::Currency.find(public_send(instance_currency_name_with_postfix))
+              elsif instance_currency_name_with_postfix.present? &&
+                     respond_to?(instance_currency_name_with_postfix) &&
+                     Money::Currency.find(public_send(instance_currency_name_with_postfix))
 
                 Money::Currency.find(public_send(instance_currency_name_with_postfix))
               elsif self.class.respond_to?(:currency)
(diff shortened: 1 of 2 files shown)


semmons99:
@antstorm could you guide them on tests?

antstorm:
@sslotnick good spot!

In your test you probably want to stub out the `currency_column` configuration. Here's the example that fails:

```
it "does not crash when currency column postfix not present" do
  allow(MoneyRails::Configuration).to receive(:currency_column) { { postfix: nil } }

  expect(product.reduced_price.currency).to eq(Money::Currency.find(:lvl))
end
```

P.S. Why would you set the currency column postfix to be nil?

sslotnick:
Thanks! Will get those updated today.

We have some tables with multiple money fields that are always in the same currency and only wanted 1 currency column. Per the comments in the configuration file for the amount_column field (https://github.com/RubyMoney/money-rails/blob/master/lib/generators/templates/money.rb#L30), we specify a full column name for the currency and do not specify a postfix at all. In hindsight, a workaround would have been to specify both a column name and postfix in the config and let the postfix get ignored. That might be why other people haven't encountered this problem.

sslotnick:
Test case added

semmons99:
Thanks!

PR #340, opened 2015-06-28
Update monetize dependency

antstorm:
Also, since `monetize` lives under the same roof, maybe it's worth omitting the patch level from the version dependency — `~> 1.3`. What do you guys think?


antstorm:
Ok, so failing mongoid tests indicate different in behaviour between AR and mongoid extensions. I'll see if this can be unified.

semmons99:
No activity in in over 4 weeks. Closing.

PR #347, opened 2015-07-22
Checking for a specific error to be raised in tests

antstorm:
Hi, 

In order for those tests to make more sense we need to check for a specific error. Otherwise any error that might occur inside of those blocks will cause the test to pass even `NoMethodError`.

This should also fix few warnings that appear while running the tests.


semmons99:
awesome

PR #350, opened 2015-07-26
Fix RuboCop lint offenses

agrimm:
Fix most of the offenses reported by RuboCop when run with the `--lint` option.

This was done so that when this gem is used while warnings are turned on (`RUBYOPT="-w"`), this gem produces fewer warnings.

The gem will still produce warnings about uninitialized instance variables being retrieved by monetizable.rb

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -14,7 +14,7 @@ end
 
 APP_RAKEFILE = File.expand_path("../spec/dummy/Rakefile", __FILE__)
 
-load 'rails/tasks/engine.rake' if File.exists?(APP_RAKEFILE)
+load 'rails/tasks/engine.rake' if File.exist?(APP_RAKEFILE)
 
 require 'rake'
 require 'rspec/core/rake_task'
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -85,10 +85,10 @@ module MoneyRails
             #
             # To disable validation entirely, use :disable_validation, E.g:
             #   monetize :price_in_a_range_cents, :disable_validation => true
-            if validation_enabled = MoneyRails.include_validations && !options[:disable_validation]
+            if (validation_enabled = MoneyRails.include_validations && !options[:disable_validation])
 
               # This is a validation for the subunit
-              if subunit_numericality = options.fetch(:subunit_numericality, true)
+              if (subunit_numericality = options.fetch(:subunit_numericality, true))
                 validates subunit_name, {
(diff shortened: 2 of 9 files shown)


Rakefile
```diff
@@ -14,7 +14,7 @@ end
 
 APP_RAKEFILE = File.expand_path("../spec/dummy/Rakefile", __FILE__)
 
-load 'rails/tasks/engine.rake' if File.exists?(APP_RAKEFILE)
+load 'rails/tasks/engine.rake' if File.exist?(APP_RAKEFILE)
```
antstorm: What's the reason for this change?

lib/money-rails/active_record/monetizable.rb
```diff
@@ -85,10 +85,10 @@ def monetize(*fields)
             #
             # To disable validation entirely, use :disable_validation, E.g:
             #   monetize :price_in_a_range_cents, :disable_validation => true
-            if validation_enabled = MoneyRails.include_validations && !options[:disable_validation]
+            if (validation_enabled = MoneyRails.include_validations && !options[:disable_validation])
```
antstorm: Hmm… not sure about this one. Is it because of the assignment inside a condition?

Rakefile
agrimm: File.exists? is deprecated - it generates a warning when Ruby is run in verbose mode from Ruby 2.1.0 onwards.

lib/money-rails/active_record/monetizable.rb
agrimm: Yes. This one doesn't generate a Ruby warning if I recall correctly, but I thought I may as well fix all the RuboCop lint offenses in one go, rather than in dribs and drabs. If you don't want this as part of the commit that's fine though.

Rakefile
antstorm: Oh, didn't know that

lib/money-rails/active_record/monetizable.rb
antstorm: I think we should follow the community style guide — https://github.com/bbatsov/ruby-style-guide#no-parens-if

And it seems like this is a default option for RuboCop

agrimm: "Note that there is an exception to this rule, namely safe assignment in condition." https://github.com/bbatsov/ruby-style-guide#safe-assignment-in-condition

antstorm: Hmm, ok, fair enough

antstorm:
That looks good to go. Thanks, @agrimm!

Issue #353, opened 2015-07-31
money_without_cents method does not round correctly


casiodk:
ActionController::Base.new.view_context.money_without_cents(10.66, no_cents: true)
=> "10"

it should be rounded up to "11"

antstorm:
@casiodk correct me if I'm wrong, but `no_cents: true` actually means shows price without cents

antstorm:
Here's the line from the `money` gem — `Money.ca_dollar(599).format(:no_cents => true) #=> "$5"`

semmons99:
Closed due to inactivity.

casiodk:
But it doesnt make sense that 5.99 without cents is 5 it should be 6, or you should be able to set a rounding option

antstorm:
@casiodk think of `no_cents` as truncation and don't use it when you're trying to round the value instead. However the need for truncation without rounding in the world of money is questionable, no doubts.

casiodk:
Yes when dealing with money, it doesnt make sense, and its really annoying to have to do the correct rounding yourself. :)

casiodk:
Otherwise a great gem, hope i don´t sound mad, i´m not ;)

antstorm:
@casiodk there's always a compromise involved — some people will find `no_cents` performing rounding to be unintuitive :)

casiodk:
<%= money_without_cents(Money.new(1066), round: true) %>

=> 11

casiodk:
could this work then

antstorm:
@casiodk sure, but do we need this when we can just:

```ruby
money_without_cents(Money.new(1066).round)
```

?

PR #354, opened 2015-07-31
round value if no_cents option is true

casiodk:


casiodk:
ActionController::Base.new.view_context.money_without_cents(10.66, no_cents: true) should eq "11"

lib/money-rails/helpers/action_view_extension.rb
```diff
@@ -16,10 +16,12 @@ def humanized_money(value, options={})
         :symbol => false
       }.merge(options)
       options.delete(:symbol) if options[:disambiguate]
-
+      
       if value.is_a?(Money)
+        value = value.to_d.round.to_money if options[:no_cents] == true && value.present?
```
greggroth: The extra `value.present?` check is unnecessary since this branch is encountered only if `value.is_a?(Money)`.

lib/money-rails/helpers/action_view_extension.rb
```diff
@@ -16,10 +16,12 @@ def humanized_money(value, options={})
...
         value.format(options)
       elsif value.respond_to?(:to_money)
+        value = value.to_d.round if options[:no_cents] == true && value.present?
```
greggroth: Similarly here, I think  `value.respond_to?(:to_d)` would be more rigorous and gets to the heart of what this line cares about than `value.present?`.

greggroth:
Do you mind adding a test to reproduce your expected behavior?

antstorm:
Similar request here — https://github.com/RubyMoney/money-rails/issues/353

I think the purpose of `no_cent: true` is to get rid of cents, not to round it. Plus this logic is already implemented in the money gem, no need to redefine it here.

PR #357, opened 2015-08-16
Cents suffix

AlexVPopov:
When using the `monetize` test helper one does not need to use the *_cents* suffix on the column to be monetized. Fixed the specs and the README to reflect that.


semmons99:
thanks!

PR #364, opened 2015-08-29
Bump monetize to 1.3.0

dunyakirkali:
I had to change 3 specs.


jimpo:
I don't like this change in behavior. It is bizarre behavior to convert nil values to `Money.new(0, *default_currency*)`. Are you open to changing the behavior of the Mongoid type to serialize/deserialize nil as nil?

PR #375, opened 2015-10-06
fixed user input not preserved if wrong decimal input

bertBruynooghe:
fixed error in user input preservation


semmons99:
Thanks!

PR #377, opened 2015-10-08
Fix validator invalid_currency error message look up path

fotos:
Instead of hardcoding the I18n message path for `invalid_currency` we should let Rails (ActiveModel) look up the message in all namespaces. This allows for more flexible messages and customization of the message per model / attribute.

The lookup path and the namespaces are described in detail in the guide:
http://guides.rubyonrails.org/i18n.html#error-message-scopes

This change is backwards compatible for existing applications.


semmons99:
awesome

PR #382, opened 2015-10-19
Relax monetize dependency; trust semver.

pda:
Depend on monetize `~> 1.1`, not `~> 1.3.0`.

This was [recently bumped from "~> 1.1.0" to "~> 1.3.0"][bump] in order to permit newer versions of monetize, however that blocks older versions for no reason.

Setting "~> 1.1" rolls back the minimum version to what it was before, while permitting future v1.x releases (which should be compatible according to semver) and still blocking potentially incompatible v2.x releases.

- [ ] Revert to `monetize ~> 1.1.0` to ensure build/tests still pass.
- [ ] Relax to `monetize ~> 1.1` to permit future backwards-compatible 1.x versions.

[bump]: https://github.com/RubyMoney/money-rails/pull/364

### DIFF
--- a/money-rails.gemspec
+++ b/money-rails.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   s.require_path = "lib"
 
   s.add_dependency "money",         "~> 6.5"
-  s.add_dependency "monetize",      "~> 1.3.0"
+  s.add_dependency "monetize",      "~> 1.1.0"
   s.add_dependency "activesupport", ">= 3.0"
   s.add_dependency "railties",      ">= 3.0"
 


pda:
Looks like money-rails specs introduced in https://github.com/RubyMoney/money-rails/pull/364 also introduce a dependency on monetize's NilClass extension which was added in 1.3. So the PR that bumped to 1.3 also broke support for older versions. That should be fixed?

Alternatively, the dependency should be set to `~> 1.3` not `~1.3.0`…

semmons99:
Seems like we should get this project compat with 1.3

pda:
I thought it was, as demonstrated by https://github.com/RubyMoney/money-rails/pull/364 which was merged into master.

antstorm:
Since we're already on `monetize 1.4`, I'm closing this. Feel free to reopen.

Issue #372, opened 2015-09-25
Issue with mass assignment of multiple monetized fields


datibbaw:
I have a model that has two monetized fields:

    monetize :minimum_order_free_delivery_cents, :default_delivery_fee_cents, allow_nil: true, with_model_currency: :currency

The problem I'm facing is when my record's `new()` or `create()` is called with multiple field assignments, e.g.:

    country_id: 6
    minimum_order_free_delivery: ''
    default_delivery_fee: '20'

The first assignment will implicitly set the model currency (e.g. country 6 uses HKD in our case); the second assignment sets the model currency to nil and the third assignment sets the model currency to the default currency.

Even if I set the currency after the `new()` is done, the multiplier may be have been wrongly applied at the third assignment.

My issue would be resolved if the model currency isn't touched when the money value is set to nil, but I'm open to ideas I haven't discovered yet :)

One way I've managed to make this work for now is to add a condition in the setter:

    def currency=(currency)
      super(currency) if currency
    end

semmons99:
Please submit a PR.

joekur:
We are running into this edge case as well. Ideally we think that if it is an instance-level currency column, that it should not be set to nil if assigning a nil money. If it is a per-field currency column, assigning that to nil seems appropriate. Happy to work on a pull if you agree @semmons99

joekur:
Here's a very simple fix: https://github.com/RubyMoney/money-rails/compare/master...joekur:instance-currency-assigning-nil

semmons99:
looks great, just submit the PR :smile:

datibbaw:
@joekur Thanks a lot! :smile_cat:

PR #385, opened 2015-10-27
Don't reset currency to nil for nil-able monetized fields

joekur:
See #372 for reference


semmons99:
Thanks!

datibbaw:
Thanks @semmons99 for merging this! Any idea when we can expect 1.4.2? :grin:

semmons99:
If the PRs slow down, probably this week.

Issue #373, opened 2015-09-28
exchange_to + to_json = Stackoverflow


SDEagle:
In a rails 4 application (ruby 2.2.2 OS X) with money-rails master and money 6.6.1 (merged that PR in a fork) when you exchange into another currency and then call `to_json` on the money object you get `SystemStackError: stack level too deep` in ActiveSupport.

```ruby
Money.new(1_00, 'EUR').to_json # => fine
Money.new(1_00, 'EUR').exchange_to('USD') # => fine
Money.new(1_00, 'EUR').exchange_to('USD').to_json # => Stackoverflow
# even worse now suddenly
Money.new(1_00, 'EUR').to_json # => Stackoverflow
```

So far I did not try out if this happens too on a currently published gem version with money 6.5. We use this setup because we use a custom exchange rate store with the interface from money 6.6. Could be another source for the issue.

The actual stacktrace is sadly not really that interesting
```
Exception: SystemStackError: stack level too deep
--
   0: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/numeric/conversions.rb:125:in `block (2 levels) in <class:Numeric>'
   1: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/range/conversions.rb:13:in `to_s'
   2: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/range/conversions.rb:13:in `to_formatted_s'
   3: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/object/json.rb:134:in `as_json'

      ...

7838: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/object/json.rb:159:in `block in as_json'
7839: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/object/json.rb:159:in `each'
7840: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/object/json.rb:159:in `map'
7841: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/object/json.rb:159:in `as_json'
7842: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/object/json.rb:50:in `as_json'
7843: <rubypath>/gems/activesupport-4.2.4/lib/active_support/core_ext/object/json.rb:159:in `block in as_json'
[...]

PR #391, opened 2015-11-17
failing spec for to_json stackoverflow issue

tim3z:
see #373 - this PR contains a minimal failing test.

It uses the most recent Money version (6.6) and the AR backed store is taken straight out of the [ruby-money doc](https://github.com/RubyMoney/money#exchange-rate-stores).

### DIFF
--- a/spec/active_record/monetizable_spec.rb
+++ b/spec/active_record/monetizable_spec.rb
@@ -120,6 +120,25 @@ if defined? ActiveRecord
         expect(product).to be_valid
       end
 
+      context 'when exchanging with custom store' do
+        before do
+          @bank = Money.default_bank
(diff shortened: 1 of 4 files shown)


semmons99:
So, there has to be something in variable exchange that we're not accounting for in the new ActiveRecord based lookup. Most likely a circular lookup for something like 'USD' to 'USD' or 'EUR' to 'EUR'.

Could you try using pry to experiment with that? I know in variable exchange we make note of them automatically.

tommyalvarez:
@semmons99 this is still happening with latests version of money and money-rails and Rails 5.2

tommyalvarez:
@semmons99 See this https://github.com/RubyMoney/money-rails/issues/373#issuecomment-433653724

antstorm:
#568 provides a default `as_json` to avoid possible stack overflow

Issue #394, opened 2015-12-03
More maintainable monetize function


GCorbel:
Hi,

I forked the repo to refactor the monetize function.

[Here is the main function](https://github.com/GCorbel/money-rails/blob/master/lib/money-rails/active_record/monetizable.rb#L97-L122). It still a lot of work to do. I plan to extract objects for `defined_currency_for_method`, `define_writer_method` and `define_reader_method`.

The goal is to increase the maintainability of the code. I hope it will be easier to read. Please, let me know what you think.

This is a part of the Christmas plan I have to contribute to open-source before the 24th of December, so, happy Christmas!

GCorbel:
I continued my work. I will be ready to do a PR tonight.

PR #395, opened 2015-12-04
Refactor the monetizable function

GCorbel:
PR for issue #394 

I did a refactoring. Basically, I extracted the long monetizable function in smaller objects. This is a just a proposition. There is probably other ways to do it. Please let me know what you think. 

If you agree, I will move specs to be unitary.

Thanks for very reliable tests.

### DIFF
--- a/lib/money-rails/active_record/model_extensions/attribute_reader.rb
+++ b/lib/money-rails/active_record/model_extensions/attribute_reader.rb
@@ -1,0 +1,59 @@
+module MoneyRails
+  module ActiveRecord
+    module ModelExtensions
+        class AttributeReader
+          attr_accessor :model, :name, :args, :attribute
+          def initialize(model, attribute, args)
+            @model = model
+            @name = attribute.name
+            @args = args
+            @attribute = attribute
+          end
+
+          def call
+            # Dont create a new Money instance if the values haven't been changed.
+            result = has_changed? ? memoized_value : change_amount
+            redefine_singletons(result)
+            result
+          end
+
+          private
+
+          def change_amount
+            # If amount is NOT nil (or empty string) load the amount in a Money
+            new_amount = Money.new(old_amount, attr_currency) unless old_amount.blank?
+
+            # Cache the value (it may be nil)
+            model.instance_variable_set "@#{name}", new_amount
+          end
+
+          def old_amount
+            @old_amount ||= model.public_send(attribute.subunit_name, *args)
+          end
+
+          def has_changed?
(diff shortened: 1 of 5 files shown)


semmons99:
I'm fine doing this. But it's a very large change. You need to ensure no new edge cases or bugs have been created by it.

GCorbel:
All tests are passing. I look at some possible use cases and I don't see what can be wrong. Can you give it a try?

antstorm:
I agree with @semmons99 on this one — huge change, but no new unit tests. All the new classes need to be properly tested in isolation.

Couple of other comments:

- Use `attr_reader`s instead of `attr_accessor`s, since you're not using setters
- `ModelExtensions` namespace is very misleading, since none of the included classes are actually extending model
- It seems that you've taken the "service"-object approach for `AttributeReader`, `AttributeWriter` and `CurrencyFinder`, but `MonetizableAttribute` is quite a different one, which makes it a bit confusing. Especially considering that they all live under the same namespace
- `AttributeReader`, `AttributeWriter` and `CurrencyFinder` can definitely share a common initializer for setting `model` and `name` attributes.

GCorbel:
@antstorm I agree with test in isolation but there is no new unit tests because there is no new behaviour.

I agree for other comments and will update my code.

Do you it's useful to continue this PR or it has no chance to be merged because it's a too big change?

antstorm:
@GCorbel yes, there's no new behaviour from the external API point of view, but there are now many more moving parts internally. This actually makes the whole thing much easier to test, which is really good.

I think that the best approach here would be to take this PR and split it into multiple ones (I guess 4 is the right number here, one per each new class and as a matter of fact — per each commit), each focusing on refactoring one part and new set of tests.

If you need help with writing tests — feel free to give me a shout. And btw, great work!

rafeDraper:
Hello @antstorm, I would like to contribute to this issue, since the last interaction is almost 4 years old, I would like to know if there is any update in it since there is an actual solution already committed.  I would like to begin by splitting them as mentioned in 4, and test them accordingly. 

@GCorbel (Great work!)

antstorm:
@rafeDraper help is always appreciated!

There's a half way solution in place to reduce the complexity in some of these methods, but nothing long term. The approach I had in mind was to move most of the logic into an initializable object and reduce the logic within the model itself down to interaction with that object. It needs further digging to figure out whether that actually works.

If you wanna start working on this — I'm happy to give you early feedback, so we can align on the best solution.

rafeDraper:
@antstorm Sure, I would love to contribute and help. Let me know how I can get more feedback on it.

PR #402, opened 2015-12-20
Fix rails 5 deprecations

richardvenneman:
In a Rails 5 project with money-rails the following deprecation warning is thrown:

```
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super.
```

Apparently [this deprecation](https://github.com/rails/rails/pull/19434) has been around for a while and I fixed it based on what I've learned from this article: [Rails 5, Module#prepend, and the End of `alias_method_chain`](http://www.justinweiss.com/articles/rails-5-module-number-prepend-and-the-end-of-alias-method-chain/).

I also updated the deprecated config option `serve_static_assets` in favour of `serve_static_files`.

Please note that `Module#prepend` was introduced in Ruby 2.0. [Rails 5 requires Ruby 2.2.2 or newer](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#ruby-versions) while Rails 4 prefers Ruby 2.0 but requires 1.9.3.

--

The tests are indeed failing on 1.9.3, but I'm not sure on how we could still support 1.9.3 while using `Module#prepend`.

### DIFF
--- a/lib/money-rails/money.rb
+++ b/lib/money-rails/money.rb
@@ -2,8 +2,9 @@ require "active_support/core_ext/module/aliasing.rb"
 require "active_support/core_ext/hash/reverse_merge.rb"
 
 class Money
+  alias_method :orig_format, :format
 
-  def format_with_settings(*rules)
+  def format(*rules)
     rules = normalize_formatting_rules(rules)
 
     # Apply global defaults for money only for non-nil values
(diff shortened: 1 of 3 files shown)


semmons99:
How about trying to use vanilla `alias` or `alias_method` so this doesn't break 1.9.3 compatibility?

richardvenneman:
Thanks for the suggestion @semmons99. Tests are now passing on 1.9.3.

Issue #408, opened 2016-01-16
Wrong implicit conversion before saving model


Nghi93:
Hi,

I am using the default currency USD.
When I assign a money object to a model and save this model I get the wrong value object.
For example: 
u = User.new
u.balance = Money.new(1000, 'EUR')
u.save
u.balance 
=> Money.new(1000, 'USD')

The currency has just changed, but not the value.
To mention is that I am using a custom exchange, but #exchange_to works fine.

I think it's a weird behavior that just the currency is changed but not also the value.
Am i supposted to convert it on my own before save?

ben-av:
For me it changes back to USD even before saving. 
Any simple way to handle this?

Nghi93:
Could you give an example?

ben-av:
```
class CashOrder < ActiveRecord::Base
  monetize :money_col_cents
end


order.money_col = Money.new(100, 'EUR')
<Money fractional:100 currency:EUR>
order.money_col
#<Money fractional:100 currency:USD>
```

Nghi93:
You can set default currency to eur or try to call
order.money_cal_cents_before_type_cast

barnett:
@Nghi93 but I believe currency should be able to be dynamic to whatever currency you want to set with (without needing to change the default). We are having the same issue, as demonstrated in #432

antstorm:
Opened a PR which should fix the issue ^

@Nghi93 and @ben-av to make sure the fix works in your cases could you give me your `monetize` definitions and DB columns used?

ben-av:
Hi @antstorm,
I used the regular 
```
add_money :orders, :money_col
```

antstorm:
@ben-av right, that means that you have a currency column and my fix should work. Can you please try it out?

```ruby
gem 'money-rails', github: 'antstorm/money-rails', branch: fix-inferred-currency-column-updating'
```

ben-av:
@antstorm 
Works great, thanks!

PR #409, opened 2016-01-17
Convert correctly before save

Nghi93:
As described in this issue https://github.com/RubyMoney/money-rails/issues/408 the conversion of a money object was not correct. It just changed the currency, but not used the conversion rate.

### DIFF
--- a/spec/active_record/monetizable_spec.rb
+++ b/spec/active_record/monetizable_spec.rb
@@ -65,6 +65,17 @@ if defined? ActiveRecord
         expect(product.price_cents).to eq(3210)
       end
 
+      context 'when currency from a Money object differs from default currency' do
+        before do
+          Money.default_bank.add_rate("EUR", "USD", 0.95)
+        end
+
+        it "converts the Money object to the default currency" do
+          product.price = Money.new(1000, "EUR")
+          expect(product.price).to eql(Money.new(950, "USD"))
+        end
+      end
+
       it "assigns the correct value from a Money object using create" do
         product = Product.create(:price => Money.new(3210, "USD"), :discount => 150,
                                  :bonus_cents => 200, :optional_price => 100)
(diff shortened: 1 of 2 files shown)


spec/active_record/monetizable_spec.rb
```diff
@@ -121,18 +132,16 @@ class SubProduct < Product
...
 
         it "raises exception when a String value with hyphen is assigned" do
           expect { product.accessor_price = "10-235" }.to raise_error ArgumentError
         end
-
-        it "raises an exception if it can't change currency" do
-          expect {
-            Product.new.price = Money.new(10, "RUB")
-          }.to raise_error(MoneyRails::ActiveRecord::Monetizable::ReadOnlyCurrencyException, "Can't change readonly currency 'USD' to 'RUB' for field 'price'")
-        end
```
semmons99: why isn't this true anymore?

Nghi93: Because beforehand exchange_to is called. By doing that Money::Bank::UnknownRate will be raised instead of ReadOnlyCurrencyException.
I think this part is covered elsewhere (in the money gem?). But if you want I can rewrite this part as well.

semmons99: Please do. 

- shane

Nghi93: I have rethought this part. The code is a bit hard to read, so please correct me if I've misunderstood the lines:
The ReadOnlyCurrencyException will be raised, when a field has another currency than the assigned one? 
But because of the conversion beforehand
```
if money_currency && current_currency != money_currency.id
```
can't be true anymore, because current_currency and money_currency are always equal.
So this part can be removed?

antstorm:
I don't think this behaviour should be a part of the `money-rails`, currency conversion should be performed explicitly before updating model's value.

It (arguably) works for read-only currency scenarios, but should be explicitly configured (with default behaviour similar to the current one).

PR #422, opened 2016-02-28
Fixes #421

### LINKED ISSUE #421: Wrong currency used in model initialization if passed from form
There is a problem if the currency is passed by a form:

```
 = f.input :price
 = f.input :price_currency, collection: currency_list
```

If the parameters are assigned to the model, money-rails in the background first initiates a Money object with the default currency (in general USD) and later assigns the currency. This is a problem if the "subunit_to_unit" (defined in the money gem) of both currencies is different.

If you try for example to assign 1,000 CLP (subunit_to_unit: 1) money-rails instantiates the object with USD (subunit_to_unit: 100), hence the value is multiplied by 100. The result is that I have 100,000 CLP assigned.

I suppose the problem could be solved by considering the subunit_to_unit when changing the currency.

tangopium:


antstorm:
@tangopium thanks for the PR, it definitely needs a couple of things before it can be merged:

1. Some tests around the new functionality
2. `if memoized && memoized.cents == amount` is a duplicate condition (see line 120) and should be at least extracted out
3. `unless amount.blank?` is also used in another branch of that condition, which kinda indicates that it has to come first, otherwise there's no point for other checks

tangopium:
I fixed the issues you mentioned and a test. Hope that helps

antstorm:
@tangopium it's getting better, good work! Why do you need `tax` value in the test? It doesn't look like you're using it later.

tangopium:
No, but in the transaction model (which is the only one in the test environment which has a currency column) the method "total_cents" is defined as monetize, what implies that a numericality validator is executed on it. If the tax value is not defined the addition fails because it tries to sum "amount_cents + nil".

lib/money-rails/active_record/monetizable.rb
```diff
@@ -117,9 +117,19 @@ def monetize(*fields)
...
-              if memoized && memoized.cents == amount &&
-                memoized.currency == attr_currency
-                result =  memoized
+              if memoized && memoized.cents == amount
+                if memoized.currency == attr_currency
+                  result =  memoized
+                else
+                  amount = memoized.amount
+                  unless amount.blank?
+                    amount = amount.to_money(attr_currency)
```
antstorm: You probably wanna set `result` here instead of overriding `amount`

lib/money-rails/active_record/monetizable.rb
```diff
@@ -117,9 +117,19 @@ def monetize(*fields)
...
+                    write_attribute subunit_name, amount.cents
+                  end
+
+                  # Cache the value (it may be nil)
+                  result = instance_variable_set "@#{name}", amount
```
antstorm: No need to set instance variable if the amount is blank, it's going to be exactly the same

lib/money-rails/active_record/monetizable.rb
```diff
@@ -117,9 +117,19 @@ def monetize(*fields)
...
               # Dont create a new Money instance if the values haven't been changed.
-              if memoized && memoized.cents == amount &&
-                memoized.currency == attr_currency
-                result =  memoized
+              if memoized && memoized.cents == amount
+                if memoized.currency == attr_currency
+                  result =  memoized
+                else
+                  amount = memoized.amount
+                  unless amount.blank?
```
antstorm: You probably wanna extract this condition outside as well, because line 135 also uses it and it seems like no other action is happening if it doesn't pass

antstorm:
@tangopium sorry for a delay, I'm happy to have this merged. Can you please squash it into a single commit? Thanks!

antstorm:
@tangopium can you please squash it? I really wanna include this in the next release.

tangopium:
Sorry for the delay too ;) Just squashed it.

antstorm:
@tangopium nice one, thanks a lot! Great work!

PR #428, opened 2016-03-14
Raise ArgumentError when options[:as] attribute equals monetizable attribute

LeKristapino:
Raise an ArgumentError when options[:as] parameter has the same name as monetizable attribute e.g:
`monetize :shipping_cost, as: :shipping_cost`
to escape `Stack overflow` error.

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -53,6 +53,13 @@ module MoneyRails
             # if none of the previous is the case then use a default suffix
             if name
               name = name.to_s
+
+              # Check if the options[:as] parameter is not the same as subunit_name
+              # which would result in stack overflow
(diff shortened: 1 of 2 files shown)


lib/money-rails/active_record/monetizable.rb
```diff
@@ -104,6 +104,11 @@ def monetize(*fields)
               end
             end
 
+            # Check if the options[:as] parameter is not the same as subunit_name
+            # which would result in stack overflow
+            if name == subunit_name
```
antstorm: You definitely wanna raise this as soon as you can, maybe before the `track_monetized_attribute` call

antstorm:
Move the check higher up, squash your commits into one and I'll be happy to merge it

antstorm:
Great job!

PR #425, opened 2016-03-02
Split monetize method apart

antstorm:
Hey all,

This is my attempt at refactoring the monetize method by breaking down all dynamically generated methods into simple calls to static methods. I think that dynamic methods is a bad pattern that people keep abusing by adding more context-wrapped variables.

I've also included quite a lot of tests to make sure these newly introduced methods do exactly what we expect from them.

PR #430, opened 2016-03-16
make monetized_attributes a hash_with_indifferent_access

bsharpe:
`monetized_attributes` was getting created as a normal hash in one place and as a HWIA in another.  By making it always a HWIA, we don't care if the accessed attr in `validator` is a `symbol` or a `string.`

We noticed this when running our tests -- validation was failing because `@attr.to_s` was being used to look up the monetized_attribute but the key was a `symbol.`

### DIFF
--- a/lib/money-rails/active_model/validator.rb
+++ b/lib/money-rails/active_model/validator.rb
@@ -6,7 +6,7 @@ module MoneyRails
         @record = record
         @attr = attr
 
-        subunit_attr = @record.class.monetized_attributes[@attr.to_s]
+        subunit_attr = @record.class.monetized_attributes[@attr]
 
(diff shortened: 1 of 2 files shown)


LeKristapino:
No tests? :(

bsharpe:
ok... tried to create tests... couldn't get it to fail.

The realized it's because we had to create a completely new version of `monetize` that includes an exchange date (what is $100 CAD worth in USD a year ago.)

antstorm:
@bsharpe yeah, I agree, `monetize` method has grown too complex, but there's some work around refactoring it — #425.

Re this PR — I'm happy to merge it, but it needs some tests to make sure others know that this is intentional.

PR #431, opened 2016-03-31
Fix ruby 1.9 support

antstorm:
With the release of rails `4.2.6`, `mail` (which is a dependency of `actionmailer`) has now a looser dependency on `mime-types` (`< 4`), causing issues with ruby 1.9 (not supported by `mime-types >= 3`).

Because of loose dependency of `mail`, this PR locks down `mime-types` to `< 3` for ruby 1.9.


antstorm:
Since this is a minor change and tests are failing because of this I'm merging it without waiting for comments. Feel free to reopen if needed.

PR #434, opened 2016-04-24
Update inferred currency column. Fixes #432. Fixes #408

### LINKED ISSUE #408: Wrong implicit conversion before saving model

### LINKED ISSUE #432: *_currency fields not updating
My expectation based on the documentation is that per-attribute currency fields will update when the corresponding amount field is updated. For example:

```
class Product < ActiveRecord::Base
  monetize :amount_cents
end

p = Product.new(amount: 100)
# => #<Product amount_cents: 100, amount_currency: 'USD'>

p.amount = p.amount.exchange_to('AUD')
# => #<Product amount_cents: 65, amount_currency: 'AUD'>
```

What I see in practice is that amount_cents is updated, but amount_currency is not.
Looking at https://github.com/RubyMoney/money-rails/blob/master/lib/money-rails/active_record/monetizable.rb I can see code to update the instance_currency field on the model, but nothing that would update (say) "#{name}#{MoneyRails::Configuration.currency_column[:postfix]}"

Are my expectations valid, or is it my responsibility to update the currency field manually when I do an exchange?

antstorm:
This will use `#{name}_currency` (by default, or `currency_column[:postfix]`) column to persist currency, which so far has only been used for detecting currency and had to be updated manually.

This should fix issues #432 and #408

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -32,19 +32,6 @@ module MoneyRails
                                               ":with_currency or :with_model_currency")
             end
 
-            # Optional accessor to be run on an instance to detect currency
-            instance_currency_name = options[:with_model_currency] ||
-              options[:model_currency] ||
-              MoneyRails::Configuration.currency_column[:column_name]
-
-            instance_currency_name = instance_currency_name &&
-              instance_currency_name.to_s
-
-            # This attribute allows per column currency values
-            # Overrides row and default currency
-            field_currency_name = options[:with_currency] ||
-              options[:field_currency] || nil
-
             name = options[:as] || options[:target_name] || nil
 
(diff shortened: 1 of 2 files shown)


antstorm:
@barnett happy to help!

antstorm:
@RubyMoney/money-rails-devs need a pair of 👀 on this one

antstorm:
Ok, I'm merging this one in

PR #437, opened 2016-05-12
Add money field for rails_admin

### LINKED ISSUE #216: Add support for rails_admin
Hello,

I started using this gem since it is working with MongoID. Unfortunately integration in `rails_admin` is not that easy. I tried around for some time but couldn't find a solution which works fine with `rails_admin`.

It would be great if there was a field type for rails_admin. There is another gem called [mongoid_money_field](https://github.com/glebtv/mongoid_money_field) which [comes with rails_admin support](https://github.com/glebtv/mongoid_money_field/blob/master/lib/mongoid_money_field/rails_admin.rb) by defining a field type for it.

I would love to see the same for `money-rails`.

What do you think about my suggestion?

yurijmi:
Fixes #216 

Unfortunately, you need to specify ```:money``` field in your model config:
```ruby
rails_admin do
  field :price, :money
end
```

That's due to rails_admin implementation.

### DIFF
--- a/lib/money-rails.rb
+++ b/lib/money-rails.rb
@@ -14,3 +14,7 @@ if defined? Rails
   require "money-rails/railtie"
   require "money-rails/engine"
 end
+
+if Object.const_defined?("RailsAdmin")
+  require "money-rails/rails_admin"
+end
--- a/lib/money-rails/rails_admin.rb
+++ b/lib/money-rails/rails_admin.rb
@@ -1,0 +1,21 @@
+require 'rails_admin/config/fields'
+require 'rails_admin/config/fields/types/integer'
+require 'money-rails/helpers/action_view_extension'
+
+include MoneyRails::ActionViewExtension
+
+module RailsAdmin
+  module Config
+    module Fields
+      module Types
+        class Money < RailsAdmin::Config::Fields::Types::Integer
+          RailsAdmin::Config::Fields::Types::register(self)
+
+          register_instance_option :pretty_value do
+            humanized_money_with_symbol(value)
+          end
+        end
+      end
+    end
+  end
+end


antstorm:
@yurijmi Looking really good! Can you add a test for it as well?

yurijmi:
@antstorm oh, I don't even know how to test it properly :( And that would definitely require adding rails_admin as testing dependency.

antstorm:
@yurijmi yeah, that's true, anyways I've tried it and it works so I'm happy to merge it. Great job!

jparr:
Note: this requires RailsAdmin to be in the gem file before money-rails

antstorm:
@jparr good catch, probably worth either fixing it or at least mentioning in the README

vemv:
Nice work!

Instead of reordering gems, you can just `require 'money-rails/rails_admin'` in the rails_admin.rb initializer.

While the `list` view in RailsAdmin looks better thanks to this request, editing a money field is still unusable to non-technical folks (try telling them to enter e.g. USD$12,353 _in cents_... every day 😀)

Am I missing something? Anyone else having the same need?

From my side I have a nice generic Rails (non-RailsAdmin) solution, based in `autoNumeric.js` plus a few bits of logic. Adapting it as a RailsAdmin field seems totally possible, but _also_ would be such a large/custom solution that I couldn't reasonably contribute to money-rails with it.

antstorm:
I was always under the impression that RailsAdmin has a good pluggable architecture, allowing for a simple extension converting values from `money-rails` into whatever HTML fields are better and back…  I might be wrong :)

vemv:
It's an interesting idea, to implement it purely in RailsAdmin!

Javascript would give extra usability (like adding dots/commas as one types), but yeah a minimally-viable implementation in RailsAdmin would be far easier to get merged into money-rails.

An would-be `Field` subclass would override `parse_value` and/or such.

antstorm:
Yes, exactly! Happy to accept a PR if you're for it

PR #411, opened 2016-01-27
make humanized_money helper respect currency decimal mark and thousan…

manuelmeurer:
…ds separator

Fixes #410

PR #443, opened 2016-06-10
Revert #411

iGEL:
I completely disagree with #411 and propose to revert it.

The formatting of the number depends on the language in the real world,not on the currency. The best example is the Euro, where it previously enforce the German format on the entire Euro zone. France, Italy and Spain for example write `1 234,56 €`, not `1.234.56 €`. And I think, English speaking countries would expect `1,234.56 €`, even though there is no English speaking country who uses the Euro. In return, we Germans also write `$ 1.234,56`, not `$ 1,234.56`. As I see it, you can't show currencies on a site in German and French now without always passing the params anymore.

The question is, whether the currency should have this information at all. Even though it comes from `money`, which might be used without `i18n`, I think it should be remove for the same reason.

/cc @manuelmeurer, @semmons99, @antstorm

### DIFF
--- a/lib/money-rails/helpers/action_view_extension.rb
+++ b/lib/money-rails/helpers/action_view_extension.rb
@@ -11,23 +11,19 @@ module MoneyRails
         options = { :symbol => options }
       end
 
-      unless value.is_a?(Money)
-        if value.respond_to?(:to_money)
-          value = value.to_money
-        else
-          return ''
-        end
-      end
-
       options = {
-        :no_cents_if_whole   => MoneyRails::Configuration.no_cents_if_whole.nil? ? true : MoneyRails::Configuration.no_cents_if_whole,
-        :symbol              => false,
-        :decimal_mark        => value.currency.decimal_mark,
-        :thousands_separator => value.currency.thousands_separator
+        :no_cents_if_whole => MoneyRails::Configuration.no_cents_if_whole.nil? ? true : MoneyRails::Configuration.no_cents_if_whole,
+        :symbol => false
       }.merge(options)
       options.delete(:symbol) if options[:disambiguate]
 
-      value.format(options)
+      if value.is_a?(Money)
+        value.format(options)
+      elsif value.respond_to?(:to_money)
+        value.to_money.format(options)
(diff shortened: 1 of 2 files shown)


antstorm:
@iGEL that's a very good point. What you said definitely makes perfect sense. Feels like there should be a separate set of rules for each country, merged with default currency rules.

@manuelmeurer what are your thoughts on this?

manuelmeurer:
I agree, didn't know the Euro (and other currencies?) actually have different formatting in different countries.

iGEL:
@antstorm From my point of view, the money gem doesn't need to know which country uses which decimal rule. Number formatting is not as easy as it might sound, for example Asian countries like China & Korea usually separate by 4 instead of 3 digits, like `1,2345.67` instead of `12,345.67`, see [Chinese_numerals](https://en.wikipedia.org/wiki/Chinese_numerals#Whole_numbers). The money gem could assume the English rules and allow overwriting them globally or per call. It sounds like a I18n concern (the concept, not the gem) and could be handled by a separate gem (like the i18n gem). 

TL;DR: I'd remove that knowledge from the money gem as well.

@manuelmeurer Thanks. I hope I don't sounded to harsh, forgot about the negative bias of written communication. :wink:

manuelmeurer:
@iGEL It's all good. 😄 Actually @semmons99 [had the same concern](https://github.com/RubyMoney/money-rails/pull/411#issuecomment-175771935) early on but I didn't understand what he meant.

antstorm:
@iGEL yeah, that sounds like `money-i18n` or something to me. I remember quite a lot of questions about formatting and it seems like everyone has it's own rules to it.

@manuelmeurer since it was your commit in the first place I want to make sure you're happy with reverting this and there are ways to get similar behaviour with an extra helper method or something like that.

manuelmeurer:
I just checked in my app and setting the locale correctly actually solves my problem mentioned in #411. So I'm good with reverting this! 👍

antstorm:
Great, I'm merging this then!

iGEL:
Thanks, guys!

Issue #384, opened 2015-10-23
Assigning to nil now returns zero-valued money object


joekur:
Just switched to latest master from 1.4.1. Found I had a test failure and tracked it down to a change in behavior when setting a monetized column to nil. It now returns a 0 value money object.

Before (1.4.1):

``` ruby
MyModel.new(price: nil).price # => nil
```

Master:

``` ruby
MyModel.new(price: nil).price # => #<Money fractional:0 currency:USD>
```

Was this intentional?

joekur:
I see this is caused by a difference in `nil.to_money`, which used to be an undefined method but now returns 0 money.

joekur:
So the monetize gem changed in 1.3.0: https://github.com/RubyMoney/monetize/blob/master/CHANGELOG.md#130

Related issue: https://github.com/RubyMoney/monetize/issues/47

I suppose this was indeed intended in Monetize. It was a bit of a breaking change (at least in my app) and took a little while to track down that it was due to a change in a dependency.

Also worth noting that this behavior is not the case if you are using `allow_nil: true`, in which case it will happily leave the monetized column as nil.

antstorm:
@joekur that's a good point. It certainly doesn't feel right to be using `allow_nil: true` in order to prevent this.

joekur:
I guess personally I would be fine with the behavior that we had before: setting to nil never defaults the value to 0. Do you agree? I'd be happy to work on a PR if so.

antstorm:
@joekur sorry for no reply, yes it definitely feels wrong — setting monetized attribute to nil should reset amount & currency. I'd be happy to review and merge a PR for this one.

joekur:
I think there's a scenario if you have more than one monetized field sharing a currency column. We would not want the currency field to be overwritten in that case. In fact I dusted off my fork and remembered I had made a branch to fix this, which is broken whether `allow_nil` is true or not.

I imagine it's possible to handle resetting the currency field if it's a shared currency column only if all dependent money fields are nil, but that gets a little complicated.

mstate:
I had the reverse situation where I wanted allow_nil due to nil data provided on input that I didn't want to translate to 0 cents in the database (too many fields and places to make the change).  But, for calculations, I didn't want to have to force a to_money on every attribute read.  I added the following to a config initializer to make nil cents values in the database be read as 0.to_money.
`
# config/initializers/money-rails-nil-to-zero.rb
# Force nil values to be zero money on read.

[...]

PR #446, opened 2016-07-01
Do not convert Mongoid money fields from nil to zero.

jimpo:
[This change](https://github.com/RubyMoney/monetize/pull/42) to monetize 1.3.0 extended NilClass to respond to `:to_money` and return a 0-valued Money object.

This is a breaking change in money-rails, as documented in https://github.com/RubyMoney/money-rails/issues/384. Whenever assigning a Mongoid Money field to nil, it will convert it to a 0-valued Money object. The behavior of money-rails was explicitly changed in https://github.com/RubyMoney/money-rails/pull/364.

For the record, I think the change to `monetize` should get reverted, but at least making `money-rails` preserve its old behavior is essential.

### DIFF
--- a/lib/money-rails/mongoid/money.rb
+++ b/lib/money-rails/mongoid/money.rb
@@ -37,6 +37,7 @@ class Money
           object.symbolize_keys!
         end
         ::Money.new(object[:cents], object[:currency_iso]).mongoize
+      when object.nil? then nil
       when object.respond_to?(:to_money) then
         begin
           object.to_money.mongoize
--- a/lib/money-rails/mongoid/two.rb
+++ b/lib/money-rails/mongoid/two.rb
@@ -21,6 +21,7 @@ class Money
         :cents        => object.cents.is_a?(BigDecimal) ? object.cents.to_s : object.cents,
         :currency_iso => object.currency.iso_code
       }
+    when object.nil? then nil
     when object.respond_to?(:to_money)
       begin
         serialize(object.to_money)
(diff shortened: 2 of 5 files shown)


antstorm:
@jimpo I totally agree with that approach — this should not be decided for you, instead it should follow Rails-approach, which is: `nil` stays `nil` until you specify a `default` value for the given field. I think it fits here nicely as well.

But I'd like to hear from other people using this feature. Especially the motivation for introducing this change in the first place.

jimpo:
Bump. What would be required to get this merged?

antstorm:
@ct-clearhaus can you please comment on this one? Mainly interested in the reasoning behind adding `to_money` to `NilClass`.

antstorm:
@jimpo since very few people use `monetize` without `money-rails` it makes more sense to change it instead of overriding that behaviour in `money-rails`.

antstorm:
@jimpo what is the reason for limiting Rails to `< 5.0`?

jimpo:
When I opened the PR, this gem wasn't passing tests on Rails 5. If Rails 5 support has been added, I'm happy to remove the constraint.

antstorm:
@jimpo yeah, it's now running fine on Rails 5. Please rebase with master and remove that constraint.

Since there's no input on this issue I'm happy to merge it because it makes sense. The actual implementation might change later if we drop `to_money` from `NilClass`, but the test is there and I'll keep an eye on it.

jimpo:
OK, rebased and tests are passing. Thanks for following up on this @antstorm, it's a huge help for us!

antstorm:
@jimpo no worries! Make sure to ping me if you have any other issues or PRs

hosh:
Bump: When is the next release with this fix expected to land?

antstorm:
@hosh just pushed `1.7.0` to rubygems

PR #451, opened 2016-08-22
fix test suite locally

anujbiyani:
I was having trouble running `rake spec:all` due to a few issues:

1. task was called `rails5` in `rake spec:rails` but it was actually named `rails50`
  * rename `rails5` to `rails50` everywhere
1. tests kept running against the wrong environment
  * solved by wrapping execution in `Bundler.with_clean_env`
  * I use `rvm` which might be why I have failures and you don't; I've tested my solution on `rbenv`, too, though and it works on both environments
2. rails 5 tests would blow up with `ActiveRecord::NoEnvironmentInSchemaError`
  * solved by ensuring the DBs are removed so that rails 5 doesn't try and look for a table that isn't created in earlier versions of Rails

I'm not sure how this will behave on Travis, we'll see I guess =)

I'm still getting mongo failures locally (specific tests are red), haven't looked into it yet. Might just ignore it for now if Travis doesn't complain.


antstorm:
@anujbiyani that's looking really good. Thank you!

PR #457, opened 2016-10-10
ADD: using a lambda for loading default_currency

mroach:
Makes life in multi-tenant applications easier!

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,19 @@ MoneyRails.configure do |config|
 end
 ```
 
+If you need to set the default currency on a per-request basis, such as in a
+multi-tenant application, you may use a lambda to lazy-load the default currency
+from a field in a configuration model called `Tenant` in this example:
+
+```ruby
+# config/initializers/money.rb
+MoneyRails.configure do |config|
+
+  # set the default currency based on client configuration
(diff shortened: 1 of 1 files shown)


antstorm:
@mroach nice one! I wasn't even aware that you could do that! :) Thanks a lot for the PR!

penhlenh:
@mroach Yes thanks for the PR. This is really nice. But could you help provide an example of how your `Tenant.current` class method might look like?

mroach:
@penhlenh It's using [acts_as_tenant](https://github.com/ErwinM/acts_as_tenant). Under the hood, that's using thread-level variables to track a "current" tenant.

In the years since making this comment, I've stopped using `default_currency` and I want to stop using `acts_as_tenant.
Having a default currency set can hide problems in your code, such as a place where currency should have been set to an explicit value but wasn't. I would rather raise an error about currency not being set than have the wrong currency be set. 100 EUR is a lot different than 100 JPY.

I'm also moving away from acts_as_tenant since it's hacking-up Rails scopes and it can be too easy to accidentally evade it. Like doing `SomeModel.unscoped` removes the tenant scoping.

penhlenh:
Ok, thanks for the explanation @mroach

PR #462, opened 2016-12-21
Clarify readme snippet for t.monetize

emilebosch:
Took me 15 minutes to figure out i was using `t.money` instead of `t.monetize` in rails 4. Since `t.money` works in rails 4 it confused the heck out of me, not understanding why it wouldn't set the currency properly.


antstorm:
@emilebosch great change, thanks!

PR #464, opened 2016-12-30
Ruby 2.4.0 support

antstorm:


oliverklee:
Does this need a rebase after #466 (or it this PR now not needed anymore)?

antstorm:
@oliverklee #466 was merged into this branch, so it doesn't need a rebase. The only thing that's preventing me from merging it with `master` straight away is the fact that we're now raising different errors (`Monetize::ParseError` instead of `ArgumentError`) and that's potentially a breaking change for some users.

It probably makes sense to keep the old approach… because otherwise you'd wanna a subclass of `MoneyRails::Error`to be raised, you don't wanna be exposed to any internal dependencies.

consti:
👍 just saw this! I believe this would resolve #469 for me.

rsempe:
@antstorm What do you mean by "keeping the old approach"?
How can I help you to merge this?

antstorm:
@rsempe by old approach I mean raise `ArgumentError` instead of the `Monetise::ParseError`. Yeah, if you could make a PR for that — would be great!

antstorm:
@rsempe I've opened a PR (#471) that would raise `MoneyRails::Error` instead of the original error, preventing from exposing our inner dependencies to the users. In case there's a need to make these more specific — client code should still work because the new errors will be subclasses of `MoneyRails::Error`.

emaiax:
Nice! I hope this can be merged soon.

antstorm:
@emaiax yes, it is happening!

emaiax:
@antstorm awesome! :shipit:

PR #466, opened 2017-01-09
Ruby 2.4.0 support and upgrade dependencies

rsempe:
- Ruby 2.4 support
- Upgrade Money dependency from 6.7 to 6.8.1
- Upgrade Monetize dependency from 1.4.0 to 1.6.0

==

The rescue of `ArgumentError` is removed in favor of `Monetize::ParseError`.

ArgumentError is only raised if value is not an instance of Numeric.
https://github.com/RubyMoney/monetize/blob/master/lib/monetize.rb#L117

But this is only called if value is a Numeric.
https://github.com/RubyMoney/monetize/blob/master/lib/monetize.rb#L62

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,8 +48,6 @@ matrix:
       gemfile: gemfiles/rails4.gemfile
     - rvm: 2.4.0
       gemfile: gemfiles/rails41.gemfile
-    - rvm: 2.4.0
-      gemfile: gemfiles/rails42.gemfile
 notifications:
   recipients:
     - andreas@aloop.org
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,20 @@ matrix:
       gemfile: gemfiles/rails50.gemfile
     - rvm: 2.1.8
       gemfile: gemfiles/rails50.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/mongoid2.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/mongoid3.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/mongoid4.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/rails3.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/rails4.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/rails41.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/rails42.gemfile
 notifications:
   recipients:
     - andreas@aloop.org
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,6 @@ gemfile:
   - gemfiles/rails50.gemfile
 matrix:
   exclude:
-    - rvm: 1.9.2
-      gemfile: gemfiles/mongoid3.gemfile
-    - rvm: 1.9.2
-      gemfile: gemfiles/rails4.gemfile
     - rvm: 1.9.3
       gemfile: gemfiles/rails50.gemfile
     - rvm: 2.0.0
--- a/gemfiles/rails42.gemfile
+++ b/gemfiles/rails42.gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'rails', '~> 4.2.0'
+gem 'nokogiri', '~> 1.6.8.1'
 
 platforms :jruby do
   gem "activerecord-jdbc-adapter"
--- a/gemfiles/rails42.gemfile
+++ b/gemfiles/rails42.gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rails', '~> 4.2.0'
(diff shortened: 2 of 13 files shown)


antstorm:
@rsempe can you please rebase it against the `ruby-2-4-0` branch and change the base for this PR?

.travis.yml
```diff
@@ -30,16 +30,26 @@ gemfile:
...
+    - rvm: 2.4.0
+      gemfile: gemfiles/mongoid4.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/rails3.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/rails4.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/rails41.gemfile
+    - rvm: 2.4.0
+      gemfile: gemfiles/rails42.gemfile
```
antstorm: Are these all incompatible with ruby 2.4.0? ^

gemfiles/rails42.gemfile
```diff
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'rails', '~> 4.2.0'
+gem 'nokogiri', '~> 1.6.8.1'
```
antstorm: What is the reason for this change?

spec/dummy/db/schema.rb
```diff
@@ -41,19 +41,19 @@
...
 
-  create_table "transactions", force: :cascade do |t|
+  create_table "transactions", :force => true do |t|
     t.integer  "amount_cents"
     t.integer  "tax_cents"
     t.string   "currency"
-    t.datetime "created_at"
-    t.datetime "updated_at"
+    t.datetime "created_at",            :null => false
+    t.datetime "updated_at",            :null => false
```
antstorm: These seem like accidental changes ^ can you please exclude that file?

gemfiles/rails42.gemfile
rsempe: 14 days ago, a new version of nokogiri (1.7) has been released (https://github.com/sparklemotion/nokogiri/commit/c5e8e124a5266d0c2779dcd79ced8204d88bf16d). 

This new version does not support a Ruby version <= 2.1 
Here is the error we had previously on Travis: `Gem::InstallError: nokogiri requires Ruby version >= 2.1.0.`.

.travis.yml
rsempe: Only Rails 4.2 and 5.0 are still supported (http://guides.rubyonrails.org/maintenance_policy.html).

I've updated the PR to also run tests on Rails 4.2 and Ruby 2.4.

All modern Rubies ship JSON as part of stdlib, an error occurs when installing `json` gem provided by `activesupport` on older versions.

It's not possible to fix version of json to `2.0` in Gemfile because ActiveSupport is requiring `json (>= 1.7.7, ~> 1.7)`.

rsempe:
@antstorm :+1:

antstorm:
@rsempe thanks! Couple of this still unclear to me:

1. What's the reason behind changes to `rails42.gemfile`? Is rails lagging behind ruby 2.4.0 release?
2. Why did you exclude most of the older gemfiles in `.travis.yml`? Does that mean that rails < 4.2 doesn't support ruby 2.4.0?

rsempe:
@antstorm Trying to do my best, but not sure it's really clear.. 

1- Changes in `rails42.gemfile`

- `gem 'nokogiri', '~> 1.6.8.1'`
14 days ago, a new version of nokogiri (1.7) has been released (sparklemotion/nokogiri@c5e8e12).

This new version does not support a Ruby version <= 2.1
https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.md#backwards-incompatibilities

I think in the future it could be nice to also end support for Ruby < 2.1 in this project.

Here is the error we had previously on Travis: `Gem::InstallError: nokogiri requires Ruby version >= 2.1.0`.
https://travis-ci.org/RubyMoney/money-rails/jobs/190311486#L197

- `gem 'rails', github: 'rails/rails', branch: '4-2-stable'`

Ruby 2.4 ships JSON as part of stdlib, an error occurs when installing `json` gem provided by activesupport on `4.2.7.1` (latest version).

`An error occurred while installing json (1.8.3), and Bundler cannot continue.`

It's not possible to fix version of json to 2.0 in Gemfile because ActiveSupport is requiring json (~> 1.7)

It's only removed for the moment on the `4-2-stable` branch thanks to this commit: https://github.com/rails/rails/commit/89a532dfaf5910891ea3acd77d36fbaa4cb8fd0b.

I've added a comment in the Gemfile to explain this.

2- Ruby exclusions in `.travis.yml`

Only Rails 4.2 and 5.0 are still supported (http://guides.rubyonrails.org/maintenance_policy.html).
So yes, for me Rails < 4.2 doesn't support Ruby 2.4.0.

We have the same issue with `json` gem here.

Ruby 2.4 ships JSON as part of stdlib, an error occurs when installing json gem provided by activesupport on older versions.

RomainOliva:
+1

antstorm:
@rsempe thank a lot for explaining this, makes perfect sense!

PR #468, opened 2017-01-12
Add humanized_money_with_iso helper

### LINKED ISSUE #467: Helper to use currency ISO code instead of symbol?
Hi! I don't think this is possible currently, but is there a helper method to display a currency amount with the ISO code? For example `USD 100` instead of `$100` as generated by `money_without_cents_and_with_symbol`.

If not I can submit a PR to add one if it would be helpful. There might be some localization issues though, since in different languages the same amount would be either `USD 100` or `100 USD` according to https://en.wikipedia.org/wiki/ISO_4217#Position_of_ISO_4217_code_in_amounts. Thanks!

rolodato:
Closes #467.

Note that `Money` has no methods or settings to place the ISO code as a prefix/suffix depending on the locale: https://github.com/RubyMoney/money/issues/600. The suggested workaround in the README is as follows:

```ruby
m = Money.new('123', :gbp) # => #<Money fractional:123 currency:GBP>
m.format( symbol: m.currency.to_s + ' ') # => "GBP 1.23"
```

If supporting suffixes is something that you would like, let me know and we can figure out a good way to support it that's easy to integrate with locales as well. Otherwise let me know and I can update the README/docs before merging.

Thanks!

### DIFF
--- a/spec/helpers/action_view_extension_spec.rb
+++ b/spec/helpers/action_view_extension_spec.rb
@@ -50,6 +50,14 @@ describe 'MoneyRails::ActionViewExtension', :type => :helper do
     it { is_expected.to include Money.default_currency.symbol }
   end
 
+  describe '#humanized_money_with_iso' do
+    subject { helper.humanized_money_with_iso Money.new(12500) }
+    it { is_expected.to be_a String }
+    it { is_expected.not_to include Money.default_currency.decimal_mark }
+    it { is_expected.to include Money.default_currency.iso_code }
+    it { is_expected.not_to include Money.default_currency.symbol }
+  end
+
   describe '#money_without_cents' do
     let(:options) { {} }
     subject { helper.money_without_cents Money.new(12500), options }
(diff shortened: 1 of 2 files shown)


antstorm:
@rolodato good change, thanks! So yes, it makes perfect sense to put this in a README for better visibility.

spec/helpers/action_view_extension_spec.rb
```diff
@@ -50,6 +50,14 @@
     it { is_expected.to include Money.default_currency.symbol }
   end
 
+  describe '#humanized_money_with_iso' do
+    subject { helper.humanized_money_with_iso Money.new(12500) }
+    it { is_expected.to be_a String }
+    it { is_expected.not_to include Money.default_currency.decimal_mark }
+    it { is_expected.to include Money.default_currency.iso_code }
+    it { is_expected.not_to include Money.default_currency.symbol }
```
antstorm: I think instead of having these ^^ 4 specs it's better to have one checking for the exact output match. Otherwise adding extra symbols to your new method will still result in this test passing.

antstorm: Same goes for all the other specs you've added

PR #470, opened 2017-01-21
money-rails without rails

brlo:
Hello and thanks for gem.
`require "money-rails/railtie"` and `require "money-rails/engine"` will fail because I don't have Rails and ::Rails::Railtie, ::Rails::Engine are not defined, but I'm using money-rails pretty long just with ActiveRecord by running `MoneyRails::Hooks.init`

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,12 @@ $ rails g money_rails:initializer
 There, you can define the default currency value and set other
 configuration parameters for the rails app.
 
+Without Rails in rack-based applications, call during initialization:
+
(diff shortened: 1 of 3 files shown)


brlo:
similar situation here https://github.com/kaminari/kaminari/issues/796

antstorm:
@brlo that's a really good change, I like that! Can you please also update the README with an example of how this can be used outside Rails?

antstorm:
@brlo looks good, thanks!

PR #471, opened 2017-01-23
Raise MoneyRails::Error intead of the original error

antstorm:


lib/money-rails.rb
```diff
@@ -5,6 +5,7 @@
 require "money-rails/money"
 require "money-rails/version"
 require 'money-rails/hooks'
+require 'money-rails/errors'
```
rsempe: What about `money-rails/error` instead and using singular form for the filename `lib/money-rails/error.rb`?

I think it could be more coherent with `MoneyRails::Error`, no?

antstorm: Yeah, I thought of that, but this would be a place to put other errors in as well… similar to https://github.com/RubyMoney/monetize/blob/master/lib/monetize/errors.rb. Sounds sane?

rsempe: Makes sense - Thanks for the explanation :+1:

PR #481, opened 2017-04-26
Mention Rails 3/4 difference w.r.t. migration

jellymann:
The README mentions the difference between Rails 3 and 4 with regards to `change_table`, but not regarding the single-line `add_money` helper.

This PR rectifies this.


antstorm:
@jellymann good shout, thank you!

PR #483, opened 2017-05-09
Add use_bank method to use it when you need to load money objects from active_record with custom bank conversion factors

zlahham:
This is to allow the usage of custom currency stores to be loaded per User sessions.

An example would be to use the following in a certain controller

```ruby
  around_action :currency_store

  def currency_store
    Money.use_bank(Money::Bank::VariableExchange.new(MyCustomStore.new)) do
      # Rate will be loaded from the current_user setup.
      Money.add_rate("USD", "EUR", 0.5)
      yield    
    end
  end
```

### DIFF
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -140,6 +140,15 @@ describe "configuration" do
         end
       end
     end
+  end
 
+  describe 'modifying default bank from memory' do
+    it 'creates new banks when needed' do
+      old_bank = Money.default_bank
+      bank = Money::RatesStore::Memory.new
+
+      Money.with_bank(bank) {}
+      expect(Money.default_bank).to eq(old_bank)
+    end
   end
 end
--- a/README.md
+++ b/README.md
@@ -359,6 +359,30 @@ t = Transaction.new(:amount_cents => 2500, :currency => "CAD")
 t.amount == Money.new(2500, "CAD") # true
 ```
 
+####  Assigning Currencies on the go
+
+If you would like to assign different Currency stores(banks) on the go you can use the following method:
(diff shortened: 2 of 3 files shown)


zlahham:
@antstorm Will do sometime early next week!

zlahham:
@antstorm Done :smile:

spec/configuration_spec.rb
```diff
@@ -140,6 +140,15 @@
...
       end
     end
+  end
+
+  describe 'modifying default bank from memory' do
+    it 'creates new banks when needed' do
+      old_bank = Money.default_bank
+      bank = Money::RatesStore::Memory.new
 
+      Money.with_bank(bank) {}
```
antstorm: You definitely wanna make sure that the block was called and the new bank is set correctly within it

antstorm:
While I can definitely think of other useful applications for this method I wonder if that's any different from just setting the `default_bank` to the one you need? Which should work due to Rails' request-based nature…

spec/configuration_spec.rb
zlahham: Will do!

zlahham:
@antstorm That's what I initially tried to do. In my case, I was setting a different bank for different Instances of the same model. So if you try to access `default_bank` it would get stuck on the first one that was set.

This was the cleanest way I could think of 😄 

I took inspiration from how the `use_timezone` method works in Rails.

antstorm:
Oh, I see, it's probably because `default_bank` is being set on the `Money` class directly. Can you please add a check that the block is called and is using the correct bank inside of it? I'll merge it then.

Issue #399, opened 2015-12-09
add_monetize and locale


nicooga:
For some reason, the migration helper `add_monetize` generates a column "<attr_name>_centavos" ("centavos" is spanish for "cents"). I don't know if this is expected behavior, but it was not expected for me and I couldn't find some one with the same "error".

Yes, I set my locale to es-AR

``` ruby
    config.i18n.available_locales = ["es-AR", "es"]
    config.i18n.default_locale = "es-AR"
    config.i18n.enforce_available_locales = false
    config.i18n.fallbacks = ["es-AR", "es", "en"]
```

... but I don't see the point, is a convention to keep variable names in english.

For the time being I resorted to just create the columns without using the macro, and with a new migration since I don't want to rollback on staging.

``` ruby
class SolveMigrationProblemWithMoneyOnTeachers < ActiveRecord::Migration
  def change
    remove_monetize :teachers, :hour_price

    add_column :teachers, :hour_price_cents, :integer, default: 0
    add_column :teachers, :hour_price_urrency, :string, default: "ARS"
  end
end
```

Just wanted to point this out as it seemed strange to me.

Cheers

rizidoro:
same here , but when a set to Brazilian Portuguese... Did you find why?

rizidoro:
So, digging into the `money` gem, it uses the `Subunit` whuch in `centavo` in BRL:
https://github.com/RubyMoney/money/blob/85159c7d8540bcac23f138372f5f64cecf2ca199/config/currency_iso.json

So this is not a issue.

nicooga:
@rizidoro still it doesn't make sense for the cents column name to be localized. The currency can change any time.

igneus:
This is even more interesting in case of default currency set to `:czk` (Czech Koruna), whose "cents" are called "haléř". money-rails isn't scared at all by the diacritics and happily creates database columns named e.g. `amount_haléřs`

antstorm:
Going through the specs I didn't find anything related to this, so I guess it is non-intentional behaviour and should be removed. Happy to review & merge a PR for this one.

antstorm:
@igneus since #425 is not getting reviewed/merged (plus it's a big change, so will probably go into the next version of `money-rails`) would be great to check out your solution for this issue and merge it.

igneus:
> @antstorm: Going through the specs I didn't find anything related to this, so I guess it is a non-intentional behaviour and should be removed.

[...]

PR #484, opened 2017-05-15
amount column postfix always '_cents'

igneus:
Until now subunit name of the default currency has been used as postfix of the amount column. This leads to various inconvenient situations, as discussed in https://github.com/RubyMoney/money-rails/issues/399

### DIFF
--- a/lib/money-rails/configuration.rb
+++ b/lib/money-rails/configuration.rb
@@ -37,7 +37,7 @@ module MoneyRails
     end
 
     def set_amount_column_for_default_currency!
-      amount_column.merge! postfix: "_#{default_currency.subunit.downcase.pluralize}" if default_currency.subunit
+      amount_column.merge! postfix: "_cents" if default_currency.subunit
     end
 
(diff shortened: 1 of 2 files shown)


lib/money-rails/configuration.rb
```diff
@@ -37,7 +37,7 @@ def register_currency=(currency_options)
     end
 
     def set_amount_column_for_default_currency!
-      amount_column.merge! postfix: "_#{default_currency.subunit.downcase.pluralize}" if default_currency.subunit
+      amount_column.merge! postfix: "_cents" if default_currency.subunit
```
antstorm: I guess this method isn't needed anymore since it's redefining an already present value

igneus:
@antstorm Thanks for the review.

I deleted the obsolete method and read the whole README carefully. It doesn't mention the effect of default currency on column naming, so no change was necessary.

antstorm:
@igneus thanks for a great change, I will release it as `1.9.0` later in the week.

PR #486, opened 2017-05-24
Update monetize dependency 1.7.0

JulienItard:
@antstorm Can i PR new version of money-rails after or need more feature before new version ? Thanks !


antstorm:
@JulienItard I've got some ideas for the `1.9.0`, so I'll try to get them merged in before releasing it. You can still use `master` in your projects if you need the new dependencies.

Also since it's a minor change and doesn't affect `money-rails` API I might as well post this as `1.8.1`.

PR #490, opened 2017-07-05
Create memoized amount using Money.new

shioyama:
We had an issue upgrading money-rails where a record with a money value would get a cents value that was 100 times larger than it should be because it was first memoized as JPY, then converted to USD using `to_money` (which interprets the amount as dollars instead of cents).

This fixed the issue so I'm just posting to get feedback. I'm not very familiar with the internals of money-rails and why you would use `to_money` instead of `Money.new` here, but there is a drastic difference when you're dealing with currencies with and without subunits.

### DIFF
--- a/spec/active_record/monetizable_spec.rb
+++ b/spec/active_record/monetizable_spec.rb
@@ -810,14 +810,13 @@ if defined? ActiveRecord
           reduced_price = product.read_monetized(:reduced_price, :reduced_price_cents)
           product.reduced_price_cents = 100
 
-          expect(product.read_monetized(:reduced_price, :reduced_price_cents)).not_to eq(reduced_price)
+          expect(product.read_monetized(:reduced_price, :reduced_price_cents)).to eq(Money.new(100, reduced_price.currency))
         end
 
         it "resets memoized attribute's value if currency has changed" do
           reduced_price = product.read_monetized(:reduced_price, :reduced_price_cents)
-          product.reduced_price_currency = 'CAD'
-
-          expect(product.read_monetized(:reduced_price, :reduced_price_cents)).not_to eq(reduced_price)
+          product.reduced_price_currency = 'JPY'
+          expect(product.read_monetized(:reduced_price, :reduced_price_cents)).to eq(Money.new(reduced_price.cents, 'JPY'))
         end
       end
 
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -193,7 +193,7 @@ module MoneyRails
           if memoized.currency == attr_currency
             result = memoized
           else
-            memoized_amount = memoized.amount.to_money(attr_currency)
+            memoized_amount = Money.new(amount, attr_currency)
             write_attribute subunit_name, memoized_amount.cents
             # Cache the value (it may be nil)
             result = instance_variable_set("@#{name}", memoized_amount)


antstorm:
@shioyama thank you for the PR! While I know this is a small change it might lead to potential bugs, so you'll need to add new specs for this. Basically you wanna test the case when currency changes to the one with a different subunit-to-unit ratio. This should make it easier to discuss the underlying issue.

The main different between using `Money.new` and `.to_money` is that the first one will just use the value it's given, which the other will try to convert the amount to the new currency's subunit-to-unit resulting in a different value, which is then used to update the `_cents` attribute.

shioyama:
@antstorm Thanks for the feedback! Yes certainly I imagine a spec will be required. I just wanted to first see if the specs would pass on Travis, and to see if you had any feedback on the change. I'm not 100% sure I understand the use of the memoized value here, and I am aware that this could be a very critical part of the gem (we had factor of 100 errors in our specs when upgrading), so I definitely want to tread carefully here.

The fact that the specs pass though would indicate to me that if this is an important change, it is not currently tested sufficiently (even without the change).

spec/active_record/monetizable_spec.rb
```diff
@@ -810,14 +810,13 @@ class SubProduct < Product
...
+          expect(product.read_monetized(:reduced_price, :reduced_price_cents)).to eq(Money.new(100, reduced_price.currency))
         end
 
         it "resets memoized attribute's value if currency has changed" do
           reduced_price = product.read_monetized(:reduced_price, :reduced_price_cents)
-          product.reduced_price_currency = 'CAD'
-
-          expect(product.read_monetized(:reduced_price, :reduced_price_cents)).not_to eq(reduced_price)
+          product.reduced_price_currency = 'JPY'
+          expect(product.read_monetized(:reduced_price, :reduced_price_cents)).to eq(Money.new(reduced_price.cents, 'JPY'))
```
shioyama: I've updated the existing spec to use a currency with a different subunit-to-unit ratio. This spec now only passes with the code change I've introduced, however a different spec is now failing.

shioyama:
@antstorm I've updated an existing spec to check a currency (JPY) with a different subunit-to-unit value. This spec only passes with the code change, *however* another spec fails:

```ruby
it "assigns the correct value from params" do
  params_clp = { amount: '20000', tax: '1000', currency:  'CLP' }
  product = Transaction.create(params_clp)
  expect(product.valid?).to be_truthy
  expect(product.amount.currency.subunit_to_unit).to eq(1)
  expect(product.amount_cents).to eq(20000)
end
```

The `product.amount_cents` here ends up being `2_000_000` instead of `20_000`. The currency here, 'CLP', is another one with an unusual subunit-to-unit ratio of 1 (like JPY).

It seems in this case that the currency is `nil` at the time when the `amount` and `tax` are being assigned, and this results in the currency defaulting to `EUR`, the amounts get interpreted as euro values (rather than eurocent values), and then when the currency is converted to CLP these are multiplied by 100.

shioyama:
Never mind, I now understand what's going on. The `amount` field on e.g. `Transaction` above should be interpreted as either dollars or cents depending on the currency. In this case this PR is not appropriate, so I'll close it.

antstorm:
@shioyama sorry for a delay. Did you figure out why all the tests have passed on your initial implementation? Did the code change when you added a new unit test?

shioyama:
@antstorm I think the test coverage is not completely covering the code, since the one-line change I made (which I now realize is not correct) passed all specs.

antstorm:
@shioyama thanks, I'll make sure to add a spec to cover that use case

Issue #491, opened 2017-07-12
Money amount interpreted as cents pre v1.6.2 if default currency was cents-based, but not after


shioyama:
I'm not sure if this is a bug before or after v1.6.2, but it's definitely something that should be flagged.

This change: https://github.com/RubyMoney/money-rails/compare/v1.6.1...v1.6.2 had a drastic effect on how money fields are interpreted depending on the value of the default currency.

I've created a repository and some specs to demonstrate this: https://github.com/degica/money-rails-test

See these spec results: https://travis-ci.org/degica/rails-money-test/builds/252703835

Specs are run with v1.6.1 and v1.6.2, and with default currency set to JPY and USD in context blocks.

In 1.6.1, with default currency set to "JPY", this is what we see (and expected):

```ruby
payment = Payment.new(amount: 100, currency: "USD")
payment.amount.cents
#=> 100

payment = Payment.new(amount: 100, currency: "JPY")
payment.amount.cents
#=> 100
```

So everything is cents-based. This happens to work well for us, and I assumed this was the way things worked. However, I didn't realize, but apparently if you default currency is "USD", this is what you saw previous to 1.6.2:

```ruby
payment = Payment.new(amount: 100, currency: "USD")
payment.amount.cents
#=> 10000

payment = Payment.new(amount: 100, currency: "JPY")
payment.amount.cents
#=> 10000
```

which seems completely crazy to me. Everything (including JPY, which are not dollars) is being treated as if it's dollars. (This I guess was the issue raised in #421.)

As of v1.6.2, however, this has changed and now the `amount` is interpreted as either a cent amount or dollar amount depending on the currency.

```ruby
payment = Payment.new(amount: 100, currency: "USD")
payment.amount.cents
#=> 10000

payment = Payment.new(amount: 100, currency: "JPY")
payment.amount.cents
#=> 100
```

I guess this makes more sense, but there does not seem to be any indication that this change was conscious (no clear mention of it anywhere, no major version change, etc.)

We luckily caught this but you can imagine that it very well could have had a huge impact on our money calculations. I'm amazed this has not been flagged earlier since the change happened more than a year ago.

I previously tried to "fix" this problem in #490 but then realized that the code is perhaps not wrong, since I do find the change as of v1.6.2 more internally consistent. However I do think there should be a warning somewhere! Can you confirm my interpretation here is correct?

Thanks!

shioyama:
[...]

PR #492, opened 2017-07-13
Add option to universally treat amounts as cents

shioyama:
This is an idea for deatling with #491.

I added an option, `treat_amounts_as_cents`, which is `false` by default but if set results in all amounts being treated as cents, so if the option is *off*:

```ruby
Payment.new(amount: 100, currency: "USD").cents
#=> 10000
```

whereas if it's *on*:

```ruby
Payment.new(amount: 100, currency: "USD").cents
#=> 100
```

This in fact was how things worked prior to v1.6.2 if you had a default currency that was cents-based (e.g. JPY). But in #421 it was pointed out that this had some strange results, such as converting 100 as an amount for a cents-based currency into 10000 if the subunit-to-unit ratio of the default currency is 100.

Obviously making conversion different depending on the default currency is a bad idea. As of v1.6.2, the interpretation of the amount depends on the currency, which is more consistent.

However, in many cases it is better to just treat *all* amounts as cents. This is in fact what RubyMoney itself does:

```ruby
Money.new(100, "USD").cents
#=> 100
```

I'm not sure if what I've done here is the best way to do this, so for now it's more of a proof of concept to discuss, and to see if specs pass on Travis.

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -193,7 +193,12 @@ module MoneyRails
           if memoized.currency == attr_currency
             result = memoized
           else
-            memoized_amount = memoized.amount.to_money(attr_currency)
+            memoized_amount =
+              if MoneyRails::Configuration.treat_amounts_as_cents
+                Money.new(amount, attr_currency)
+              else
+                memoized.amount.to_money(attr_currency)
+              end
             write_attribute subunit_name, memoized_amount.cents
             # Cache the value (it may be nil)
             result = instance_variable_set("@#{name}", memoized_amount)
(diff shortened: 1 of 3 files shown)


antstorm:
Ok, so unfortunately this doesn't really solve the problem, only complicates the code. You can get exactly the same behaviour by using the subunit column directly. Assuming in your case it's `cents` you can do `Payment.new(cents: 100, currency: 'USD').cents` giving you back `100`.

You can achieve the same result with `Payment.new(amount: Money.new(100, 'USD'))`.

And because there are at least 2 other ways to set cents directly I don't think there's a need for option that will enable another way of doing it.

shioyama:
I agree about not merging this. Like I said, I wanted to see what it would take to make this work.

However:

> You can get exactly the same behaviour by using the subunit column directly. Assuming in your case it's cents you can do Payment.new(cents: 100, currency: 'USD').cents giving you back 100.

That's not so easy to do in a large code base, where we have been setting the money field from user params assuming that integers get converted to cents (which they used to). The behaviour I'm after here is that a money field (`amount` in our case) takes either an integer/string or a money, and if integer/string converts to money as cents.

The subunit column (`amount_cents`) *does not accept* a money object, so it's not a simple swap-out of one for the other.

```ruby
Payment.new(amount: Money.new(100, 'USD'))
```

This works, but would require changes to all places where models are created/updated. But I'm leaning toward doing this since it seems the safest solution.

I'll close this and move discussion back to #491.

PR #511, opened 2017-11-21
Return early from reading monetized attribute if nil amount and allow_nil is set

Eldwick:
## Issue:
For a monetized attribute that is nil and is allowed to be nil, a corresponding currency is always attempted to be read. This causes issues because the `currency_for` method might call the `Money.default_currency` which for some multi-tenant applications could throw an error if called outside of a specific context.

## Solution:
If the monetized attribute is nil and is allowed to be nil, immediately return nil without attempting to read the corresponding currency.

### DIFF
--- a/spec/active_record/monetizable_spec.rb
+++ b/spec/active_record/monetizable_spec.rb
@@ -870,18 +870,9 @@ if defined? ActiveRecord
           Money.default_currency = original_default_currency
         end
 
-        context "when allow_nil options is not set" do
-          let(:options) { {} }
-
-          it "attempts to read the fallback default currency" do
-            expect(default_currency_lambda).to receive(:read_currency).and_return("USD")
-            subject
-          end
-        end
-
         context "when allow_nil options is set" do
           let(:options) { { allow_nil: true } }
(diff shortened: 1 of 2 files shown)


Eldwick:
@antstorm Thanks! I have added the tests for this use case.

spec/active_record/monetizable_spec.rb
```diff
@@ -854,6 +854,39 @@ class SubProduct < Product
...
+
+        around(:each) do |example|
+          service # Instantiate instance which relies on Money.default_currency
+          original_default_currency = Money.default_currency
+          Money.default_currency = -> { default_currency_lambda.read_currency }
+          example.run
+          Money.default_currency = original_default_currency
+        end
+
+        context "when allow_nil options is not set" do
```
antstorm: I think you can remove this test case since if the `allow_nil` is not set it should follow the regular path and is probably tested somewhere else

Eldwick:
@antstorm Removed the duplicate test

antstorm:
@Eldwick thank you for pushing this through!

PR #514, opened 2018-01-24
Add test helper for with_model_currency

keithbro:
So that in your Rails model specs you can do something like:

````
it { is_expected.to monetize(:sale_price_amount)
  .as(:sale_price)
  .with_model_currency(:sale_price_currency_code) }
````

### DIFF
--- a/lib/money-rails/test_helpers.rb
+++ b/lib/money-rails/test_helpers.rb
@@ -81,21 +81,21 @@ module MoneyRails
 
       def test_allow_nil
         if @allow_nil
-          @actual.send(@money_attribute_setter, "")
-          @actual.send(@money_attribute).nil?
+          @actual.public_send(@money_attribute_setter, "")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/money-rails/test_helpers.rb [hunk 1]
```diff
@@ -95,6 +101,14 @@ def test_currency_iso
         end
       end
 
+      def test_currency_attribute
+        if @currency_attribute
+          @actual.send(@money_attribute).currency == @actual.send(@currency_attribute)
```
antstorm: Can you please change `send`s to `public_send`s?

keithbro: I've replaced all instances of `@actual.send` with `@actual.pubic_send`. Is that what you wanted, @antstorm? What is the reasoning behind this?

keithbro:
Added documentation, as requested 👍

PR #523, opened 2018-04-09
Update money to 6.11

swrobel:
CHANGELOG: https://github.com/RubyMoney/money/blob/master/CHANGELOG.md#6110

Do we need to be so restrictive with minor versions?

### DIFF
--- a/money-rails.gemspec
+++ b/money-rails.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
 
   s.require_path = "lib"
 
-  s.add_dependency "money",         "~> 6.10.0"
+  s.add_dependency "money",         "~> 6.11.0"
   s.add_dependency "monetize",      "~> 1.7.0"
   s.add_dependency "activesupport", ">= 3.0"
   s.add_dependency "railties",      ">= 3.0"


antstorm:
@swrobel the failing tests in TravisCI suggest that we probably should be :)

swrobel:
Not SemVer, huh?

antstorm:
@swrobel it is, however it seems like `money-rails` depends on a private method from `Money`, which is pretty bad and should be fixed. Semver unfortunately doesn't protect against similar money patches

pedantic-git:
I have a fix for that issue, which also removes the use of the private method. But there's another issue with money 6.11 to do with using ',' as a decimal point in some currencies. Just seeing if I can resolve that too.

My version of money.rb is pasted below in the meantime if it helps!

```ruby
require "active_support/core_ext/module/aliasing.rb"
require "active_support/core_ext/hash/reverse_merge.rb"

class Money
  class <<self
    alias_method :orig_default_formatting_rules, :default_formatting_rules

    def default_formatting_rules
      rules = orig_default_formatting_rules || {}
      defaults = {
        no_cents_if_whole: MoneyRails::Configuration.no_cents_if_whole,
        symbol: MoneyRails::Configuration.symbol,
        sign_before_symbol: MoneyRails::Configuration.sign_before_symbol
      }.reject { |_k, v| v.nil? }

      rules.reverse_merge!(defaults)

      unless MoneyRails::Configuration.default_format.nil?
        rules.reverse_merge!(MoneyRails::Configuration.default_format)
      end
      rules
    end
  end
end
```

antstorm:
Closing this in favour of #525

PR #525, opened 2018-04-10
Make money-rails compatible with money 6.11

pedantic-git:
This PR should fix the issues with #523.

I must admit I'm not very familiar with Money so I would appreciate code review in case this is wrong, but the tests do pass with these changes (on rails50 at least).

Changes I've made:

1. Override Money 6.11's public accessor for `default_formatting_rules`. You may prefer to assign this directly rather than override it? This replaces the monkeypatch that was in there before.
2. Check for a `Money` type coming back from `before_type_cast` when checking to see if something is stringy. This apparently fixes some I18n issues that the tests flagged up.


pedantic-git:
Thanks so much!

PR #527, opened 2018-04-16
chore: update README with wider explanation of the attribute currency option and its values

### LINKED ISSUE #526: Can I make a PR to update the README for the ':with_currency' option of 'monetize'?
As I was reading through the code I've noticed that `Attribute Currency (:with_currency)`, besides specifying a currency code as it is currently written, has an option to also accept a [proc, lambda (or any Ruby object that responds to a method named `call`)](https://github.com/RubyMoney/money-rails/blob/master/lib/money-rails/active_record/monetizable.rb#L282) as it value. As far I can see it is also tested in a couple of situations ([one example](https://github.com/RubyMoney/money-rails/blob/master/spec/active_record/monetizable_spec.rb#L1021) out of many)

My humble opinion is that this is a really cool feature, and while I understand that the gem/engine is not huge, still it would be better if it is elaborated more (with small examples of course) on this.

ristovskiv:
close #526

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ object using EUR as their currency, instead of the default USD.
 #### Attribute Currency (:with_currency)
 
 By passing the option ```:with_currency``` to the ```monetize``` macro call,
-with a currency code as its value, you can define a currency in a more granular
+with a currency code (symbol or string) or a callable object (object that responds to the method ```call```) that returns a currency code, as its value, you can define a currency in a more granular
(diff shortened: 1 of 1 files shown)


antstorm:
Looks great, thank you for adding this 👍

ristovskiv:
@antstorm thank you and if you have anything else that you need some work on, I am willing to try to help? :smile:

antstorm:
@ristovskiv that's great, really appreciate it!

PR #537, opened 2018-08-09
Update money to 6.12

swrobel:


antstorm:
I've published monetize 1.9.0, which has relaxed money version (~> 6.12), so minor updates should be easier now

Issue #524, opened 2018-04-10
undefined method `normalize_formatting_rules' for #<Money fractional:1000 currency:EUR>


damienlethiec:
Hello there,

I have just created a new Rails 5.2 app and, when I try to use the helper `humanized_money_with_symbol` with a correct Money object, I get the following error:

`undefined method `normalize_formatting_rules' for #<Money fractional:1000 currency:EUR>`

Any clue of the reason? Any fix?

Thanks a ton :)

PedroPauloML:
Same error

antstorm:
Which version of money are you using? This seems to be caused by the new version on money, however `money-rails` depends on `~> 6.10.0`, which should not include the latest release

antstorm:
This has been fixed by #525. Also new version of the gem has been pushed to RubyGems — https://rubygems.org/gems/money-rails (version 1.11.0). Please upgrade and the issue should go away.

vladprotsevich:
Hi! Did u find a way how to solve that issue? I updated my gem to 1.13.3 but that problem not disappear

PR #546, opened 2018-10-21
update version of gem added to Gemfile

mbugowski:
if user added gem 'money-rails', '~>1' they may get error mentioned in https://github.com/RubyMoney/money-rails/issues/524 version which was fixed later (in 1.11.0)


antstorm:
@mbugowski thank you for fixing this! 👍

PR #549, opened 2018-10-25
Replace currency_as_string with currency.to_s

Jcambass:
This replaces all usages of `currency_as_string` with `currency.to_s`.
`currency_as_string` currently causes some deprecation warnings.


antstorm:
@Jcambass looks good, thank you!

PR #552, opened 2018-11-09
Drop older rails/mongoid support

antstorm:
We should only be supporting latest rails/mongoid releases for each major version


FunkyloverOne:
Hey @antstorm, don't forget to also update README :) (It still says there's only Mongoid 2 and 3 support)

antstorm:
@FunkyloverOne good shout, will do 👍

Issue #397, opened 2015-12-07
Get only cents part from money


iBublik:
Is there any built-in way to get only cents part of the money? Thanks in advance :)

amartinezg:
Hi @iBublik 

You have a method called `.cents` for each Money Object.

`# 10.00 USD`
`money = Money.new(1000, "USD")`
`money.cents     #=> 1000`
`money.currency  #=> Currency.new("USD")`

See: https://github.com/Rubymoney/money#usage

Hope it helps

Cheers

iBublik:
Hi @amartinezg. Thanks for your reply. I think I've provided too little info so you misunderstood me. What I wantis to get 50 (excactly cents without dollars) from following example:
```ruby
money = Money.new(1050, 'USD')
humanized_money(money) #=> $10.50
# following method probably have another name or doesn't' exist, it's here just to explain what I want
only_cents_without_dollars(money) #=> 50
```
Currently I'm using `subunit_to_unit` to get this value, but maybe there's pretty built-in way.

antstorm:
@iBublik I would suggest raising this in [money repo](https://github.com/RubyMoney/money). Right now it's being done like this:

```ruby
unit, subunit = fractional.abs.divmod(currency.subunit_to_unit)
```

PR #553, opened 2018-11-14
Add money_only_cents helper method

carlomartinucci:
Following issue #397, I think this helper method may be useful.

I'm sorry if I break some convention or standard.

### DIFF
--- a/lib/money-rails/helpers/action_view_extension.rb
+++ b/lib/money-rails/helpers/action_view_extension.rb
@@ -50,12 +50,9 @@ module MoneyRails
     end
 
     def money_only_cents(value)
-      if value.is_a?(Money)
-      elsif value.respond_to?(:to_money)
-        value = value.to_money
-      else
-        return '00'
-      end
+      return '00' unless value.respond_to?(:to_money)
+
+      value = value.to_money
 
       format "%0#{value.currency.exponent}d", (value % value.currency.subunit_to_unit).cents
     end
--- a/lib/money-rails/helpers/action_view_extension.rb
+++ b/lib/money-rails/helpers/action_view_extension.rb
@@ -48,5 +48,16 @@ module MoneyRails
     def money_without_cents_and_with_symbol(value)
       money_without_cents(value, symbol: true)
     end
+
+    def money_only_cents(value)
+      if value.is_a?(Money)
+      elsif value.respond_to?(:to_money)
+        value = value.to_money
+      else
+        return '00'
+      end
+
+      format "%0#{value.currency.exponent}d", (value % value.currency.subunit_to_unit).cents
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/money-rails/helpers/action_view_extension.rb
```diff
@@ -48,5 +48,16 @@ def money_without_cents(value, options={})
     def money_without_cents_and_with_symbol(value)
       money_without_cents(value, symbol: true)
     end
+
+    def money_only_cents(value)
+      if value.is_a?(Money)
```
antstorm: There's probably no need for this empty branch

carlomartinucci: thank you, i didn't realize that `money.to_money == money`.

carlomartinucci:
@antstorm thank you, I added the spec (hope it's good, I followed the other ones).
My specific use case was for styling reason, so `.format(html_wrap: true)` and a little more css would do.

Another use case could be, for instance, if you want integer part and cents part in different td of the same row in a table, or similar complex settings.
Another hypothetical use case could be to hide the real amounts in a list but make them distinguishable, similar to how credit card numbers work.

antstorm:
@carlomartinucci the spec looks good, but can you also verify the exception case when a non-money object was given. It might also make sense to check that a single digit is prefixed with zero, just in case.

carlomartinucci:
@antstorm I hope I understood. I added test for a monetizable object (a number) and for less than 10 cents. Do you want me to add also a test for an object that does not respond to `to_money`? (e.g. `true`?)

spec/helpers/action_view_extension_spec.rb
```diff
@@ -74,6 +74,31 @@
     it { is_expected.not_to include "00" }
   end
 
+  describe '#money_only_cents' do
+    let(:monetizable_object){ Money.new(125_00) }
+    subject { helper.money_only_cents monetizable_object }
+    it { is_expected.to be_a String }
+    it { is_expected.not_to include Money.default_currency.decimal_mark }
+    it { is_expected.not_to include Money.default_currency.symbol }
+    it { is_expected.to include "00" }
```
antstorm: This should be `is_expected.to eq "00"` (instead of `include`), otherwise result of `12500` would also pass this spec

antstorm:
@carlomartinucci yeah, that's exactly right, a spec case that takes `return '00' unless value.respond_to?(:to_money)` path, otherwise all the specs that you have will actually respond to `to_money`

spec/helpers/action_view_extension_spec.rb
carlomartinucci: good catch, thanks

carlomartinucci:
@antstorm I added the spec for `false` (the only non-monetizable object i could think of, beside `true` ;) ) and also removed some useless `it`'s, since `is_expected.to eq ...` covers it all. :)

antstorm:
I wonder why is that failing on CI…

carlomartinucci:
I really have no idea... Should I do anything?

PR #557, opened 2018-12-03
Mongoid: add a guard clause for "blank form" input

FunkyloverOne:
e.g. when given a hash with :cents & :currency_iso keys as blank strings.
This will also work well with my (not yet pushed to GitHub) special RailsAdmin Mongoid Money field.
Also minor refactoring for `Money.mongoize` class method.

### DIFF
--- a/lib/money-rails/mongoid/money.rb
+++ b/lib/money-rails/mongoid/money.rb
@@ -28,25 +28,12 @@ class Money
     # Takes any possible object and converts it to how it would be
     # stored in the database.
     def mongoize(object)
-      case
-      when object.is_a?(Money) then object.mongoize
-      when object.is_a?(Hash) then
-        if object.respond_to?(:deep_symbolize_keys!)
-          object.deep_symbolize_keys!
-        elsif object.respond_to?(:symbolize_keys!)
-          object.symbolize_keys!
-        end
-        ::Money.new(object[:cents], object[:currency_iso]).mongoize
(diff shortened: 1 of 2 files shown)


FunkyloverOne:
BTW, for me `rake spec:mongoid2` task fails, because it kinda tries to use ActiveRecord in those specs, I'm not sure, did not investigate it further.

FunkyloverOne:
Thanks! I believe that it might be not wanted behavior for some existing apps... Maybe we should have a configuration parameter fot it?

FunkyloverOne:
@antstorm you know what, it's not a breaking change, it's rather a bug fix, because without it - it would raise `Money::Currency::UnknownCurrency` exception.
And now there will be no exceptions, and value will be treated as `nil`, which is more expected outcome than an exception IMO.
This is actually why I made this PR - I believe that such exception is unexpected behaviour (at least I did not expect it when encountered it), kinda buggy... :)

FunkyloverOne:
BTW, do you think it would be better to extract all the mongoid-related stuff into separate gem? like "rails-money-mongoid"?

FunkyloverOne:
@antstorm how about a version bump, and writing the changes down to changelog? :)

antstorm:
Sure, will release a new version 👍

antstorm:
@FunkyloverOne new version `1.13.1` should now be available

PR #558, opened 2018-12-06
Using Money.locale_backend instead of Money.use_i18n and providing downward compatibility

Joerg-Seitz:
**Problem:**
The latest version of the money gem deprecates the use of `Money.use_i18n` and suggests using `Money.locale_backend`. For that reason we updated our configuration as follows:
```
Money.use_i18n = false        # We did this to avoid deprecation warnings see https://github.com/RubyMoney/money/issues/828
Money.locale_backend = :i18n
```
Since money-rails currently does not check for `Money.locale_backend` this caused the thousands_separator and the delimiter to be sourced from the currency locale. Now this normally should not cause problems as long as the thousands_separators and delimiters are identical.
But we were using an english language locale and a euro currency locale, which do not share the same delimiters and operators and caused the following validation to fail.

```
def thousand_separator_after_decimal_mark
  thousands_separator.present? && decimal_pieces.length == 2 && decimal_pieces[1].include?(thousands_separator)
end
```

**Solution:**
To fix this behaviour I introduced `use_backend_locale?` that checks if `Money.backend_locale` is set and only calls `Money.use_i18n` as a fallback to provide downwards compatibility.
Additionally this should fix deprecation warnings for people who set up their project using `Money.locale_backend`.

**Tests:**
I added tests for the new behaviour but did not adjust all the existing tests to use Money.locale_backend. Let me know if this is necessary

### DIFF
--- a/lib/money-rails/active_model/validator.rb
+++ b/lib/money-rails/active_model/validator.rb
@@ -62,16 +62,20 @@ module MoneyRails
 
       def decimal_mark
         character = currency.decimal_mark || '.'
-        @_decimal_mark ||= Money.use_i18n ? I18n.t('number.currency.format.separator', default: character) : character
+        @_decimal_mark ||= use_backend_locale? ? I18n.t('number.currency.format.separator', default: character) : character
       end
 
       def thousands_separator
         character = currency.thousands_separator || ','
-        @_thousands_separator ||= Money.use_i18n ? I18n.t('number.currency.format.delimiter', default: character) : character
+        @_thousands_separator ||= use_backend_locale? ? I18n.t('number.currency.format.delimiter', default: character) : character
       end
 
       def symbol
-        @_symbol ||= Money.use_i18n ? I18n.t('number.currency.format.unit', default: currency.symbol) : currency.symbol
+        @_symbol ||= use_backend_locale? ? I18n.t('number.currency.format.unit', default: currency.symbol) : currency.symbol
+      end
+
+      def use_backend_locale?
+        Money.locale_backend.present? || Money.use_i18n
       end
 
       def abs_raw_value
--- a/spec/active_record/monetizable_spec.rb
+++ b/spec/active_record/monetizable_spec.rb
@@ -767,9 +767,42 @@ if defined? ActiveRecord
             end
           end
 
+          context "when locale_backend is true" do
+            around(:each) do |example|
+              begin
+                Money.locale_backend = :i18n
+                Money.use_i18n = false
+                example.run
+              ensure
+                Money.locale_backend = nil
+                Money.use_i18n = true
+              end
+            end
+            it "validates with the locale's decimal mark" do
+              transaction.amount = "123,45"
+              expect(transaction.valid?).to be_truthy
+            end
+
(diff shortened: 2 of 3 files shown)


lib/money-rails/active_model/validator.rb
```diff
@@ -62,16 +62,20 @@ def currency
...
+        @_thousands_separator ||= use_backend_locale? ? I18n.t('number.currency.format.delimiter', default: character) : character
       end
 
       def symbol
-        @_symbol ||= Money.use_i18n ? I18n.t('number.currency.format.unit', default: currency.symbol) : currency.symbol
+        @_symbol ||= use_backend_locale? ? I18n.t('number.currency.format.unit', default: currency.symbol) : currency.symbol
+      end
+
+      def use_backend_locale?
+        Money.locale_backend.present? || Money.use_i18n
```
antstorm: For backwards compatibility you probably wanna check whether Money responds to `locale_backend`, otherwise there will always be a default `locale_backend` value

antstorm:
Looks like some specs are failing, any idea why?

PragTob:
:man_shrugging: I'd hope it's because of the reason you mentioned but not sure.. it seems like it's for rails5, mongoig 6 and mongoid 7 (the failures).

Always thsoe 2 small specs:

```
  1) MoneyRails::ActiveRecord::Monetizable.monetize fails validation with the proper error message if money value is invalid decimal
     Failure/Error: expect(product.errors[:price].size).to eq(1)
     
       expected: 1
            got: 2
     
       (compared using ==)
     # ./spec/active_record/monetizable_spec.rb:225:in `block (3 levels) in <top (required)>'
  2) MoneyRails::ActiveRecord::Monetizable.monetize fails validation with the proper error message if money value is nothing but periods
     Failure/Error: expect(product.errors[:price].size).to eq(1)
     
       expected: 1
            got: 2
     
       (compared using ==)
     # ./spec/active_record/monetizable_spec.rb:232:in `block (3 levels) in <top (required)>'
Finished in 1.76 seconds (files took 1.58 seconds to load)
266 examples, 2 failures
Failed examples:
rspec ./spec/active_record/monetizable_spec.rb:222 # MoneyRails::ActiveRecord::Monetizable.monetize fails validation with the proper error message if money value is invalid decimal
rspec ./spec/active_record/monetizable_spec.rb:229 # MoneyRails::ActiveRecord::Monetizable.monetize fails validation with the proper error message if money value is nothing but periods
The command "bundle exec rspec" exited with 1.
```

antstorm:
@PragTob yep, seems to be related to the new rails (5.2.2) release. Will fix it on `master` once I figure out what's wrong with it

antstorm:
@PragTob the `master` should be fixed now (#559). Can you please rebase your branch and run the specs again?

PragTob:
@antstorm thanks for fixing it! :) It's not my branch though but I'll ping @Joerg-Seitz :)

antstorm:
@PragTob sorry, my bad :) @Joerg-Seitz can you please take a look?

Joerg-Seitz:
Forgot about this for a bit. It should work now :)

PR #563, opened 2019-03-26
Lock sqlite3 to ~> 1.3.6, so specs run smoothly

FunkyloverOne:
With unlocked sqlite3 I had issues running specs, namely errors like this:
```
LoadError: Please install the sqlite3 adapter: `gem install activerecord-sqlite3-adapter` (can't activate sqlite3 (~> 1.3.5), already activated sqlite3-1.4.0. Make sure all dependencies are added to Gemfile.)
```

Now I can successfully run all the specs, yay!

PR #564, opened 2019-03-26
MoneyValidator: use composition over inheritance

FunkyloverOne:
It makes more sense, as we do not own `NumericValidator` code, and do not override anything.
What we actually do is simply prepare values for `NumericValidator`, and perform some checks after `NumericValidator` finish.
So current implementation shows our intentions clearer, and is safer, as we do not risk to accidentally override some private method of some further `NumericValidator` implementation.


FunkyloverOne:
All the failed checks are fixed in #563
Is it possible to re-run all the checks on top of that PR, once it's merged?

antstorm:
@FunkyloverOne I like this change! #563 is now merged so you should be able to rebase with `master` and I'll merge this one

PR #565, opened 2019-03-26
[WIP] Refactor Money for Mongoid, with few questions

FunkyloverOne:
Hey, @antstorm, please take a look.

I got a few questions there in a code (comments), and we might want to change the behavior a bit, depending on the answers.

Also, what if we got some garbage Hash, without needed keys at all for example? Currently, I believe it will end up with zero Money object, but I guess it better be `nil`, or we could also raise an error, depending on `MoneyRails.raise_error_on_money_parsing`.

Also, I think that blank strings, like `"     \n "` for example, also could be interpreted as `nil`, and again, even depending on `MoneyRails.raise_error_on_money_parsing`.

What do you think?

### DIFF
--- a/lib/money-rails/mongoid/money.rb
+++ b/lib/money-rails/mongoid/money.rb
@@ -13,16 +13,12 @@ class Money
     # Get the object as it was stored in the database, and instantiate
     # this custom class from it.
     def demongoize(object)
-      if object.is_a?(Hash)
-        if object.respond_to?(:deep_symbolize_keys)
-          object = object.deep_symbolize_keys
-        else
-          object = object.symbolize_keys
-        end
-        object.has_key?(:cents) ? ::Money.new(object[:cents], object[:currency_iso]) : nil
-      else
-        nil
-      end
+      return nil unless object.is_a?(Hash)
+
+      object = object.symbolize_keys
+      return nil unless object.has_key?(:cents)
+
+      Money.new(object[:cents], object[:currency_iso])
     end
 
     # Takes any possible object and converts it to how it would be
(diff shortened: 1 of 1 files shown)


lib/money-rails/mongoid/money.rb
```diff
@@ -33,6 +29,8 @@ def mongoize(object)
       return nil if object.nil?
       return mongoize_castable(object) if object.respond_to?(:to_money)
 
+      # QUESTION: why whould we even want to store it?
```
antstorm: You might wanna double check this, but I think this is the way Mongoid expects it to behave (I think I saw some other examples of this). I agree that it seems weird that you'd just use whatever has been given to you.

In case that's not true, I'm happy to return `nil` or raise here

lib/money-rails/mongoid/money.rb
```diff
@@ -48,23 +46,22 @@ def evolve(object)
...
-        hash.symbolize_keys!
-      end
+      hash = hash.symbolize_keys
 
       # Guard for a blank form
       return nil if hash[:cents] == '' && hash[:currency_iso] == ''
 
-      ::Money.new(hash[:cents], hash[:currency_iso]).mongoize
+      # QUESTION: what if Money::Currency::UnknownCurrency will be raised?
+      # Should it be dependent on `MoneyRails.raise_error_on_money_parsing`?
```
antstorm: That's a great question. I think it makes sense not to raise when that option is on and return `nil`. I think it used to be the behaviour at some point, not sure why it changed. Might be worth digging through commit history

antstorm:
I prefer a stricter behaviour when dealing with money so a random hash should not result in a zero money object. As for the string parsing — it is handled by `monetize` gem, which I'm extending to support "strict" parsing that will handle these use cases.

lib/money-rails/mongoid/money.rb
FunkyloverOne: Yeah, I guess it was me who actually broke it, will fix.

PR #568, opened 2019-04-04
Support JSON serialization

antstorm:
By default `as_json` serializes all instance variables, which seems like a bad idea, so this provides a minimal implementation to avoid default serialization.

It's probably a good idea to give people ability in the future to change this depending on their needs which might vary significantly.


antstorm:
Another possible solution is to implement `to_hash` instead, which ActiveSupport will fall back to in case `as_json` wasn't found

PR #570, opened 2019-04-29
Add locale_backend default + comments to template

### LINKED ISSUE #569: add locale_backend to configuration template
From reading the code, it looks like `MoneyRails.configure` delegates `locale_backend`, so you can deal with this issue>(https://github.com/RubyMoney/money#deprecation) within the `configure` block.

Can we add a comment for this setting to `lib/money-rails/configuration.rb` to make this clear to future users? Happy to submit a small PR if requested.

ditsara:
Per https://github.com/RubyMoney/money#deprecation the default behavior
of checking i18n locale and falling back to per-currency localization
will be deprecated. This results in a deprecation message when
`locale_backend` has not been set.

This commit explicitly sets `config.locale_backend` to `nil` and adds
comments to turn on Money format localization.

Fixes #569

### DIFF
--- a/lib/generators/templates/money.rb
+++ b/lib/generators/templates/money.rb
@@ -105,8 +105,6 @@ MoneyRails.configure do |config|
   # In case you don't need localization and would like to use default values
   # (can be redefined using config.default_format):
   # config.locale_backend = nil
-  #
-  config.locale_backend = nil
 
   # Set default raise_error_on_money_parsing option
   # It will be raise error if assigned different currency
--- a/lib/generators/templates/money.rb
+++ b/lib/generators/templates/money.rb
@@ -83,6 +83,31 @@ MoneyRails.configure do |config|
   #   sign_before_symbol: nil
   # }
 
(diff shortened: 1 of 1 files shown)


adrienpoly:
maybe it could be mentioned that this requires at least the version v1.13.2. I was on v1.13.1 and `locale_backend` wasn't available in that version, took me a bit of time to find out.

antstorm:
@ditsara thank you for the PR, looks great. One thing is that this is a breaking change, so would need a major version update. Do you think it makes sense to set it to `:legacy` as per `money`'s default or just not set it to anything, allowing users to pick themselves?

ditsara:
@adrienpoly @antstorm good point about the different versions of Money and it being a breaking change. In that case, I think leaving it unset probably makes the most sense. Money will then determine what to do -- for v1.13.2 it will output a deprecation message, and then the user can make the choice from there. I'll push a change to the branch.

antstorm:
@ditsara nice one, thank you! 👍

PR #583, opened 2019-10-15
Add rails 6 to gemfiles and travis config

smudge:
👋 Howdy -- should be pretty straightforward. I also added a sprocket 4.0 manifest file since tests are now pulling in that version of sprockets and it was causing them to fail. Let me know if you have any questions or would like me to make any adjustments.


antstorm:
@smudge thanks, this is great! :+1:

PR #588, opened 2019-12-10
Support editing of Money fields in Rails Admin

dvanoni:
This solves the issue of being unable to edit the cents in a Money field.

As an example, consider this config for a Money field in Rails Admin for use with database columns `price_cents` and `price_currency`:

```ruby
configure :price, :money
exclude_fields :price_cents, :price_currency
```

Attempting to edit the field results in the following error because of the `Integer` field type.

image

Changing the `Money` field type to extend `String` instead of `Integer` solves this issue.

### DIFF
--- a/lib/money-rails/rails_admin.rb
+++ b/lib/money-rails/rails_admin.rb
@@ -8,7 +8,7 @@ module RailsAdmin
   module Config
     module Fields
       module Types
-        class Money < RailsAdmin::Config::Fields::Types::String
+        class Money < RailsAdmin::Config::Fields::Types::Decimal
           RailsAdmin::Config::Fields::Types::register(self)
 
           register_instance_option :pretty_value do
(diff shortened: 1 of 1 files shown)


antstorm:
@dvanoni any reason why `String` instead of a `Float`?

dvanoni:
@antstorm I originally went with `String` because I wanted the value to be passed without any sort of transformations so that RubyMoney could handle it appropriately. However, upon looking at the Rails Admin code some more, it appears that both the `Float` and `Decimal` types don't do any transformations. I tested out with `Decimal`, and it appears to work as expected.

I decided to go with `Decimal` over `Float` because—even though they appear to behave the same way in Rails Admin—I believe the semantics of the `Decimal` type are more appropriate as we want an exact decimal value rather than a floating point approximation.

Let me know your thoughts. Thanks!

dvanoni:
@antstorm let me know if you get time to take another look at this. My team would love to be able to make use of it. Thanks!

dvanoni:
@antstorm I wanted to bump this again in hopes of getting another review if you get a chance 🤞 Thanks!

dvanoni:
Thanks for the merge!

PR #591, opened 2019-12-31
Fix validator race condition

### LINKED ISSUE #589: NameError: instance variable @_currency not defined
We have a high throughput application ~21k RPM. We are facing an issue with validator. About 1% of requests raises `NameError: instance variable @_currency not defined` at [this line](https://github.com/RubyMoney/money-rails/blob/master/lib/money-rails/active_model/validator.rb#L59).

Gemfile
`gem 'money-rails', '~>1.13'`

From the code it seems this error should never get raised, since it performs a `instance_variable_defined?` check. But somehow our error stack is filled with NameError's emanating from this line. I guess a rescue block can solve this issue temporarily for us.

antstorm:
Rails caches validator instances on model classes, which makes it unsafe to cache any model instance specific data on the validator. In a threaded environment lots of issues can happen because of this, e.g. `@record` can get changed on the validator when validating 2 different models at the same time.

This PR rewrites the validators to remove any memoization. The future plan for this code is to be replaced with `monetize` call once strict parsing is supported.

Fixes #589

### DIFF
--- a/lib/money-rails/active_model/validator.rb
+++ b/lib/money-rails/active_model/validator.rb
@@ -2,43 +2,54 @@ module MoneyRails
   module ActiveModel
     class MoneyValidator < ::ActiveModel::Validations::NumericalityValidator
       def validate_each(record, attr, _value)
-        reset_memoized_variables!
-        @record = record
-        @attr = attr
-
-        subunit_attr = @record.class.monetized_attributes[@attr.to_s]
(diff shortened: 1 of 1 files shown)


antstorm:
The failure is due to https://github.com/rails/rails/commit/892eab777c418135ce0646e91bc9ebb08a29ab9b. It has already been fixed, but the new version has not yet been pushed

PR #600, opened 2020-06-29
Fix deprecation warnings in ruby 2.7

### LINKED ISSUE #598: Migration Helper Deprecation Warning (Ruby 2.7.1)
When using the migration helper, `t.monetize :field`, and you run `db:migrate`, it raises a deprecating warning.

`/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/money-rails-1.13.3/lib/money-rails/active_record/migration_extensions/table_pg_rails4.rb:8: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`

I'm running ruby 2.7.1 with rails 6.0.3.1

szechyjs:
Fixes #598

Ruby 2.7 requires the use of the double splat operator `**`, unfortunately this was not added until v2.0.0. That being said I also removed support for ruby 1.9.3 (its been EOL for 5 years, so I think it's about time).

Edit: also drop ruby v2.0, v2.1 & v2.2.

### DIFF
--- a/lib/money-rails/active_record/migration_extensions/schema_statements_pg_rails4.rb
+++ b/lib/money-rails/active_record/migration_extensions/schema_statements_pg_rails4.rb
@@ -11,7 +11,7 @@ module MoneyRails
 
         def remove_monetize(table_name, accessor, options = {})
           [:amount, :currency].each do |attribute|
-            column_present, table_name, column_name, type, = OptionsExtractor.extract attribute, table_name, accessor, options
+            column_present, table_name, column_name, type = OptionsExtractor.extract attribute, table_name, accessor, options
             remove_column table_name, column_name, type if column_present
           end
         end
--- a/lib/money-rails/active_record/migration_extensions/schema_statements_pg_rails4.rb
+++ b/lib/money-rails/active_record/migration_extensions/schema_statements_pg_rails4.rb
@@ -2,16 +2,16 @@ module MoneyRails
   module ActiveRecord
     module MigrationExtensions
       module SchemaStatements
-        def add_monetize(table_name, accessor, options={})
+        def add_monetize(table_name, accessor, options = {})
           [:amount, :currency].each do |attribute|
-            column_present, *opts = OptionsExtractor.extract attribute, table_name, accessor, options
-            add_column(*opts) if column_present
+            column_present, table_name, column_name, column_type, opts = OptionsExtractor.extract attribute, table_name, accessor, options
+            add_column(table_name, column_name, column_type, **opts) if column_present
           end
         end
 
-        def remove_monetize(table_name, accessor, options={})
+        def remove_monetize(table_name, accessor, options = {})
           [:amount, :currency].each do |attribute|
-            column_present, table_name, column_name, type, _ =  OptionsExtractor.extract attribute, table_name, accessor, options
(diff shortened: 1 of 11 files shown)


antstorm:
@szechyjs thank you for the PR. The change itself makes perfect sense and I agree that we should drop support for older rubies. This is a breaking change though, so would need a major version bump and we might wanna bundle it with dropping support for more Ruby versions (probably up to 2.3)

antstorm:
@szechyjs thank you for updating the PR 👍 Let me figure out what else might we need to include in the release. Can you meanwhile reset base to the `v2` branch?

szechyjs:
> @szechyjs thank you for updating the PR 👍 Let me figure out what else might we need to include in the release. Can you meanwhile reset base to the `v2` branch?

base is now `v2`

lib/money-rails/active_record/migration_extensions/table_pg_rails4.rb
```diff
@@ -2,16 +2,16 @@ module MoneyRails
...
+            column_present, _, column_name, column_type, opts = OptionsExtractor.extract attribute, :no_table, accessor, options
+            column(column_name, column_type, **opts) if column_present
           end
         end
 
-        def remove_monetize(accessor, options={})
+        def remove_monetize(accessor, options = {})
           [:amount, :currency].each do |attribute|
-            column_present, _, column_name, _, _ =  OptionsExtractor.extract attribute, :no_table, accessor, options
+            column_present, _, column_name, = OptionsExtractor.extract attribute, :no_table, accessor, options
```
antstorm, line 14: I'm assuming that comma after `column_name` is not needed ^

lib/money-rails/active_record/migration_extensions/schema_statements_pg_rails4.rb
```diff
@@ -2,16 +2,16 @@ module MoneyRails
...
+            column_present, table_name, column_name, column_type, opts = OptionsExtractor.extract attribute, table_name, accessor, options
+            add_column(table_name, column_name, column_type, **opts) if column_present
           end
         end
 
-        def remove_monetize(table_name, accessor, options={})
+        def remove_monetize(table_name, accessor, options = {})
           [:amount, :currency].each do |attribute|
-            column_present, table_name, column_name, type, _ =  OptionsExtractor.extract attribute, table_name, accessor, options
+            column_present, table_name, column_name, type, = OptionsExtractor.extract attribute, table_name, accessor, options
```
antstorm, line 14: Same here ^

lib/money-rails/active_record/migration_extensions/table_pg_rails4.rb
szechyjs: its needed. it throws away the other return values. instead of making an array of them.

antstorm: Does it?

```ruby
>> a, b, c = [1,2,3,4]
=> [1, 2, 3, 4]
>> p a, b, c
1
2
3
````

szechyjs: yup, your right, my bad.

PR #606, opened 2020-09-22
Add option null: for both columns at the same time in the migration

thebravoman:
Allow for passing null: true and this would apply to both columns - amount and currency
```
add_column :products, :optional_price, null: true
```


thebravoman:
I can see that some of the mongo are failing. I would have to set the mongo on my side. It might take more time because of this.

```
BUNDLE_GEMFILE='gemfiles/mongoid2.gemfile' bundle install --quiet
Bundler could not find compatible versions for gem "bundler":
  In mongoid2.gemfile:
    rails (~> 3.1) was resolved to 3.2.22.5, which depends on
      bundler (~> 1.0)

  Current Bundler version:
    bundler (2.1.4)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?

Could not find gem 'bundler (~> 1.0)', which is required by gem 'rails (~> 3.1)', in any of the sources.
rake aborted!
```
I will have to set up many different envs on my machine

thebravoman:
Some checks have failed with 
```
invalid gem: package is corrupt, exception while verifying: wrong number of arguments (given 2, expected 1) (ArgumentError) in /home/travis/.rvm/gems/ruby-head/cache/bundler-1.17.3.gem

The command "gem install bundler -v '~> 1.17'" failed and exited with 1 during .
```
I am not sure whether I could do anything about them either?

antstorm:
@thebravoman I don't think this is related to this PR, let me check what might be wrong with it

antstorm:
Fixed it in #607, try rebasing with master to get it green

thebravoman:
@antstorm checks have passed.

lib/money-rails/active_record/migration_extensions/options_extractor.rb
```diff
@@ -3,7 +3,12 @@ module ActiveRecord
     module MigrationExtensions
       class OptionsExtractor
         def self.extract(attribute, table_name, accessor, options = {})
-          default = MoneyRails::Configuration.send("#{attribute}_column").merge(options[attribute] || {})
+          # Placing the default_for_both_columns first allows for 
+          # the options of the specific column to take precendence when they are passed. 
+          default_for_both_columns = {}
+          default_for_both_columns =default_for_both_columns.merge(null: options[:null]) if options.has_key?(:null)
```
antstorm, line 9: nit: missing space after `=`

antstorm: Btw, a cleaner approach might be:

```ruby
default_for_both_columns = { null: options[:null] }.compact
```

which will filter out keys with `nil` values

antstorm (approved):
Overall looks good, left some improvement suggestions

lib/money-rails/active_record/migration_extensions/options_extractor.rb
thebravoman: Would you like me to improve it?

elliotcm:
Happy to see this, I just got stuck figuring it out myself.

lib/money-rails/active_record/migration_extensions/options_extractor.rb
semmons99: @thebravoman please add the improvements and resubmit

thebravoman: @semmons99 Ok. Thanks. I will need some time to get back, because it's been a few months since then.

README.md
```diff
@@ -137,10 +137,16 @@ monetized field, you can use the `:allow_nil` parameter:
 monetize :optional_price_cents, allow_nil: true
 
 # in Migration
+# To set only the 'amount' column as null: true
 def change
   add_money :products, :optional_price, amount: { null: true, default: nil }
 end
 
+# in Migration to set both 'amount' and 'currency' columns to null: true
+def change
```
antstorm, line 146: Might be a conflict artefact, but this is redefining the `#change` method

antstorm (changes requested):

README.md
thebravoman: You are right. It is in the example. It is another example how the change method could be used. It is not redefining as the example does not represent a definition.

I could think about moving it as a new example separate example.

semmons99:
closing until improvements are made

thebravoman:
Just to be clear
The improvements that have to be made are from

https://github.com/RubyMoney/money-rails/pull/606#discussion_r495620894

semmons99:
yes

PR #619, opened 2021-01-17
Create ruby.yml

semmons99:


antstorm:
@semmons99 thoughts on squashing PRs when merging? Quite a lot of noise in the commit messages otherwise

semmons99:
@antstorm I personally do not worry about a clean commit history. I find it is better for teaching and training to so the full creative process and not a modified end result.

antstorm:
@semmons99 it does however make it difficult to build a changelog out of commits and unfortunately makes others question credibility (speaking from experience here)

semmons99:
The first is totally valid and I muscle I have not flexed in quite some time. Thanks for reminding me. I will start considering this again when making public repo commits.

I will throw down with anyone who questions someone's credibility based on their commit history 💪 😎

antstorm:
@semmons99 thank you 👍 😃

Issue #608, opened 2020-10-01
Unexpected behaviour change between 1.11.0 and 1.12.0


wjessop:
I'm seeing an unexpected behaviour change between money-rails 1.11.0 and 1.12.0 (the behaviour introduced in 1.12.0 is present in the latest version, but I narrowed it down to this version jump). I'm not sure if this issue is in money-rails, money or monetize so opening an issue here.

These two Rails console sessions sum up the issue:

## v1.11.0

```ruby
Loading development environment (Rails 5.1.7)
[1] pry(main)> BigDecimal('0.2525e1').to_money('EUR')
=> #<Money fractional:253 currency:EUR>
[2] pry(main)> MoneyRails::VERSION
=> "1.11.0"
```

## v1.12.0

```ruby
Loading development environment (Rails 5.1.7)
[1] pry(main)> BigDecimal('0.2525e1').to_money('EUR')
=> #<Money fractional:252 currency:EUR>
[2] pry(main)> MoneyRails::VERSION
=> "1.12.0"
```

Basically, the `to_money` method does additional rounding in v.1.12.0, which unfortunately throws calculations out in our app. This is the config we're currently using:

```ruby
MoneyRails.configure do |config|
  config.default_currency = :eur
  config.rounding_mode = BigDecimal::ROUND_HALF_EVEN
  config.default_format = {
    format: "%n %u"
  }
end
```

wjessop:
Further investigation, if I monkey patch the `Monetize` module with these two methods taken from 1.7.0 the previous behaviour is restored, and I can upgrade money-rails 1.13.3.

```ruby
module Monetize
  def self.from_numeric(value, currency = Money.default_currency)
    case value
    when Integer
      from_fixnum(value, currency)
    when Numeric
      value = BigDecimal.new(value.to_s)
      from_bigdecimal(value, currency)
    else
      fail ArgumentError, "'value' should be a type of Numeric"
    end
  end

  def self.from_bigdecimal(value, currency = Money.default_currency)
    currency = Money::Currency.wrap(currency)
    value *= currency.subunit_to_unit
    value = value.round unless Money.infinite_precision
    Money.new(value, currency)
  end
end
```

antstorm:
@wjessop while this might be a change in behaviour, the value should actually be 252:

```ruby
>> BigDecimal('0.2525e1').round(2, BigDecimal::ROUND_HALF_EVEN)
=> 0.252e1
```

[...]

PR #626, opened 2021-03-02
Add warning about breaking change to rounding logic

wjessop:
[The issue](https://github.com/RubyMoney/money-rails/issues/608) describes the underlying problem in detail, but in short there's a breaking change in 1.12.0 that it would be a good idea to warn about.


semmons99 (approved):

semmons99:
thank @wjessop

PR #631, opened 2021-03-17
Add Russian translation for errors (#592)

kirillshamardin:


semmons99:
please update CONTRIBUTORS, CHANGELOG, and version, then I will merge it. thank you

config/locales/money.ru.yml
```diff
@@ -0,0 +1,4 @@
+ru:
+  errors:
+    messages:
+      invalid_currency: должнa быть корректной валютой (например '100', '5%{decimal}24', или '123%{thousands}456%{decimal}78'). Получено %{currency}
```
antstorm, line 4: @semmons99 should we fix the logic here first? The message hints at an invalid currency, but it's actually suggesting how to format the number correctly

semmons99: good call. @kirillshamardin could you work with @antstorm on fixing the error definition

kirillshamardin: @antstorm please check it now.

PR #649, opened 2022-01-04
Support validation when `monetized` is used

### LINKED ISSUE #627: `monetize`d attribute autoconverts Hash to Money but validation rejects
Today, I noticed inconsistent validation behavior when validating the numericality of a provided value. In particular, if you set an attribute to a Hash, it will be type cast to a Money object but the validation will reject it as invalid. This can be illustrated by adding the following tests to `spec/active_record/monetizable_spec.rb` at line 403

```ruby
         it "passes validation when Money is given" do
        product = Product.create(price_in_a_range: Money.new(5000, 'usd'))
        expect(product.errors[:price_in_a_range].size).to eq(0)
      end

      it "passes validation when Hash is given" do
        product = Product.create(price_in_a_range: { cents: 5000, currency:'usd'})
        expect(product.price_in_a_range).to eq(Money.new(5000, 'usd'))
        expect(product.errors[:price_in_a_range].size).to eq(0)
      end
```

The first test illustrates that assigning a Money object which is larger than 0 and less than $100 (the requirement set for `price_in_a_range`) to `price_in_a_range` validates properly.

However, the second test illustrates though that the `price_in_a_range` will properly type cast the Hash to a Money object (the first expectation passes) but `price_in_a_range` is still treated as invalid. 

I think the cause of this is in `#normalize` at [line 115](https://github.com/RubyMoney/money-rails/blob/71b4d8fca9d1a3a9dd23caf63d1c83871801b898/lib/money-rails/active_model/validator.rb#L115). If `details.raw_value` is Money object, Numeric or String, the `#to_s` on line 115 will lead to a text representation of a number. However, if you provide a Hash, you'll end up with something like this: `"{:amount=>4440, :currency=>\"usd\"}"`.

wwahammy:
Supports validation when a hash is provided to a `monetized` attribute.

Example:

Given Product:

```ruby
class Product
  monetize :price_in_a_range_cents, allow_nil: true,
                                    subunit_numericality: {
                                      greater_than: 0,
                                      less_than_or_equal_to: 10000
                                    },
                                    numericality: {
                                      greater_than: 0,
                                      less_than_or_equal_to: 100,
                                      message: "must be greater than zero and less than $100"
                                    }
end
```

the following code passes validation:

```ruby
Product.create(price_in_a_range:{amount: 4, currency: 'usd'})
Product.create(price_in_a_range:{cents: 4})
Product.create(price_in_a_range:{}) # treats it as if `nil` was passed
```
While the following would not pass:

```ruby
Product.create(price_in_a_range:{currency: 'usd'}) # treat it as if `0` was passed to cents
Product.create(price_in_range:{some_other_key: 289}) # treat it as if `0` was passed to cents
```

Closes #627

### DIFF
--- a/lib/money-rails/active_model/validator.rb
+++ b/lib/money-rails/active_model/validator.rb
@@ -30,7 +30,7 @@ module MoneyRails
         return if options[:allow_nil] && raw_value.nil?
 
         # Set this before we modify raw_value below.
-        stringy = raw_value.present? && !raw_value.is_a?(Numeric) && !raw_value.is_a?(Money)
+        stringy = raw_value.present? && !raw_value.is_a?(Numeric) && !raw_value.is_a?(Money) && !raw_value.is_a?(Hash)
 
         if stringy
           # TODO: This is supporting legacy behaviour where a symbol can come from a i18n locale,
@@ -39,6 +39,11 @@ module MoneyRails
 
           # remove currency symbol
           raw_value = raw_value.to_s.gsub(symbol, "")
+        end
+
+        if raw_value.present? && raw_value.is_a?(Hash)
+          
+          raw_value = raw_value.to_money(currency)
         end
 
         # Cache abs_raw_value before normalizing because it's used in
--- a/lib/money-rails/active_model/validator.rb
+++ b/lib/money-rails/active_model/validator.rb
@@ -42,7 +42,6 @@ module MoneyRails
         end
 
         if raw_value.present? && raw_value.is_a?(Hash)
-          
           raw_value = raw_value.to_money(currency)
         end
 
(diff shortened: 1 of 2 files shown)


lib/money-rails/active_model/validator.rb
```diff
@@ -41,6 +41,11 @@ def validate_each(record, attr, _value)
           raw_value = raw_value.to_s.gsub(symbol, "")
         end
 
+        if raw_value.present? && raw_value.is_a?(Hash)
+
```
semmons99, line 45: this is rather pedantic, but would you please remove this unneeded empty line

semmons99 (changes requested):

lib/money-rails/active_model/validator.rb
wwahammy: Heh, no problem :smile:

wwahammy:
Done!

semmons99 (approved):

semmons99:
need to figure out why test is failing, hang tight

semmons99 (approved):

semmons99:
happy to merge once tests are passing

PR #651, opened 2022-01-08
Add Rails 7 and Ruby 3.1 to CI matrix

petergoldstein:
This adds Rails 7 and Ruby 3.1 to the CI runs.  Other than adding the Rails 7 gemfile and the Ruby version to the CI configuration, the only additional change was excluding Rails 7 from being run if the Ruby version is less than 2.7.


petergoldstein:
@semmons99 Just out of curiosity, why did this get closed?

semmons99:
at the time it was broken and had no updates. happy to accept a new functional PR

petergoldstein:
@semmons99 So not sure why you say that's the case, but ok.  It was never enabled by the maintainers here (which is why this PR shows no run), and it ran fine on a fork - https://github.com/petergoldstein/money-rails/actions/runs/1670401650/jobs/2265904943 .  And rebasing and adding Ruby 3.2 continues to show it running correctly.  I'll reopen this with the rebased and updated code (since it's still on the existing branch).

PR #658, opened 2022-07-24
Add ActiveJob serializer for Money objects

sunny:
This change adds a serializer so that ActiveJob can enqueue jobs with Money arguments out of the box.

I’ve been using the same serializer in a production app for a while, and think this would help users not have to create their own serializer.

See also: [Documentation for Active Job serializers](https://guides.rubyonrails.org/active_job_basics.html#serializers).

PR #661, opened 2022-12-08
Add ActiveJob serializer for Money objects

sunny:
This change adds a serializer so that ActiveJob can enqueue jobs with Money arguments out of the box.

I’ve been using the same serializer in a production app for a while, and think this would help users not have to create their own serializer.

See also: [Documentation for Active Job serializers](https://guides.rubyonrails.org/active_job_basics.html#serializers).

(Follow-up to #658)

### DIFF
--- a/lib/money-rails/active_job/money_serializer.rb
+++ b/lib/money-rails/active_job/money_serializer.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module MoneyRails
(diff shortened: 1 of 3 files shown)


semmons99 (approved):

semmons99:
TY

Aesthetikx:
It would be totally sweet if we could get a release for this ☮️

PR #669, opened 2023-04-13
Fix currency change during init

### LINKED ISSUE #645: monetized field subunit value incorrect when currency exponent changes during construction
```ruby
class Foo < ActiveRecord::Base
  # Use USD as model level currency because it has exponent 2
  register_currency :usd

  monetize :bar_amount, as: :bar, with_model_currency: :bar_currency_code, disable_validation: true
end

f = Foo.new({
  bar: 1,
  bar_currency_code: 'JPY' # exponent 1
})
f.bar_amount # this should be 1, but instead is 100
```

Reading `bar` before saving will fix the `bar_amount`, and putting `bar_currency_code` before `bar` in the initialization hash also prevents the bug.

Demonstrated by this commit with some hacked together tests:
https://github.com/asavageiv/money-rails/commit/187139212180e051d2afbf35b57dd3c60ab9d535

I think the solution to this is to modify the subunit amount when the currency changes for the field or perhaps adding an after_initialize hook that reads the field to fix up the value. Happy to submit a PR if you can guide me on the right approach.

alan-pie:
Partially fixes #645. This works for construction but not for every change of the instance variable. I'd love more feedback on how to approach that.

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -148,6 +148,13 @@ module MoneyRails
             after_save do
               instance_variable_set "@#{name}_money_before_type_cast", nil
             end
+
+            # This hack ensures that if the currency exponent changed during
+            # initialization the subunit is updated correctly.
+            # See https://github.com/RubyMoney/money-rails/issues/645
+            after_initialize do
+              amount = public_send(name)
+            end
           end
         end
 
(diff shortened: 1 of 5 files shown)


semmons99:
hmm, this is quite the stumper. I will continue to give it some thought

semmons99:
any further thoughts on this? given the length it has been open, I am planning to close if we make no further progress

PR #677, opened 2023-08-23
Document raise_error_on_money_parsing

### LINKED ISSUE #676: Wrong currency is ignored when not using `_currency` attribute
When a model does not have a currency attribute, `monetize` does not check whether the value passed to `amount=` uses the correct currency.

e.g.

```ruby
class Payment < ApplicationRecord
  monetize :amount_cents, with_currency: :usd
end

p = Payment.new
p.amount = Money.from_cents(100, :eur)
puts p.valid? # true
puts p.amount.inspect # #<Money fractional:100.0 currency:USD>
```

The currency is silently converted from `EUR` to `USD` with no error.

TastyPi:
This config option was undocumented leading me to open #676.

Closes #676

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,12 @@ MoneyRails.configure do |config|
   #   symbol: nil,
   #   sign_before_symbol: nil
   # }
+
+  # Set whether an error should be raised when parsing money values
+  # This includes assigning to a monetized field with the wrong currency
+  # Default value is false
+  #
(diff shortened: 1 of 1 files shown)


sunny (approved):
🙏🏻

chase-stevens (approved):

semmons99:
ty

PR #688, opened 2024-02-17
Fix money_only_cents for negative money

bertiecroll:
**What/Why**
When given negative money, `money_only_cents` view helper does not return the expected subunit value due to how modulo handles negative numbers.
For example: -125 % 100 = 75

**What was done**
To fix, I've updated the calculation to use the absolute value and added a new spec to cover the use case.

### DIFF
--- a/spec/helpers/action_view_extension_spec.rb
+++ b/spec/helpers/action_view_extension_spec.rb
@@ -105,6 +105,11 @@ describe 'MoneyRails::ActionViewExtension', type: :helper do
       let(:monetizable_object){ false }
       it { is_expected.to eq "00" }
     end
+
+    context 'with a negative monetizable object' do
+      let(:monetizable_object){ Money.new(-125) }
+      it { is_expected.to eq "25" }
+    end
   end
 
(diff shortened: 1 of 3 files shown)


spec/helpers/action_view_extension_spec.rb
```diff
@@ -105,6 +105,11 @@
       let(:monetizable_object){ false }
       it { is_expected.to eq "00" }
     end
+
+    context 'with a negative monetizable object' do
+      let(:monetizable_object){ Money.new(-125) }
```
sunny, line 110: Minor presentation fix to clarify the cents part here:

```suggestion
      let(:monetizable_object){ Money.new(-1_25) }
```

sunny (approved):
Looks good to me! 🔥

bertiecroll:
Thanks for the quick review, I've added the suggested improvement 👍🏽

semmons99 (approved):

semmons99:
heads up. we have failing tests here.

semmons99:
ignore. I removed 2.6/2.7 from testing

Issue #507, opened 2017-10-13
Overriding the setter throws an error


amingilani:
My values come with a lot of weird whitespace and padding so, as recommended by the [docs](http://api.rubyonrails.org/classes/ActiveRecord/Base.html#class-ActiveRecord::Base-label-Overwriting+default+accessors) I tried to override the accessors

```
  def price=(value)
    price = value.strip
    super(price)
  end
  
  def quantity=(quantity)
    quantity = quantity.strip
    super(quantity)
  end
```

This resulted in the following error:

```
     NoMethodError:
       super: no superclass method `price=' for #<Item:0x007faf30f2f620>
       Did you mean?  price
```

Is there a workaround for this? For the moment, i'm going to strip it in the controller, but this definitely feels like something the model should be doing.

samstickland:
I think this will work if you `prepend` a module instead:

```
class MyRecord < ApplicationRecord
  module StripValues
    def price=(value)
      price = value.strip
      super(price)
    end
  
    def quantity=(quantity)
      quantity = quantity.strip
      super(quantity)
    end
  end

  prepend StripValues
```

The problem is that the setter methods are defined (via `define_method`) in your active record class itself, so there is no setter method in `ActiveRecord::Base` to call. Prepending a module however will prepend the method in *your* active record class, so when it calls super it will should call the right methods.. That is, in the example above when `MyRecord#price=` is called, it will first invoke `MyRecord::StripValues#price=` and then when that calls super it will invoke `MyRecord#price=`

amingilani:
Thank you!

achernik:
There is also another way to handle this problem:

```ruby
  alias price_from_monetize= price=

  def price=(value)
    price = value.strip
    self.price_from_monetize(value)
  end
```

rockwellll:
> There is also another way to handle this problem:
> 
> ```ruby
>   alias price_from_monetize= price=
> 
>   def price=(value)
>     price = value.strip
>     self.price_from_monetize = price
>   end
> ```

I'm sorry but im dumbfounded and cant seem to understand how this works. Could you clarify? Thanks.

dorianmariefr:
@rockwellll A common idiom is to do `alias old_price= price=` so you have an alias of the current `price=` method named `old_price=` that will be used later.

Then we override `price=`, do our custom code (the strip) and call the aliased method (the old_price= method, here named `price_from_monetize=`)

hope this helps

PR #698, opened 2024-04-24
Allow nil to be set as default currency

sirwolfgang:
This is a part of fixing the https://github.com/RubyMoney/money-rails/issues/697

PR #703, opened 2024-07-03
Define methods on dynamic module

jbennett:
Currently you cannot override the generated methods and access super ie:

```ruby
class Transaction < ApplicationRecord
  monetize :price_cents

  def price=(value)
    # stuff
   super # NoMethodError
  end
end
```

This happens because the method is defined on the class directly so their is no super method to call. A solution today is to create an alias (`alias :old_price= :price=`) before overriding the method as is recommending in issue #507.

Instead, a new module can be created, included into the class, and the methods defined on it allowing `super` to work as expected.

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -9,6 +9,8 @@ module MoneyRails
       extend ActiveSupport::Concern
 
       module ClassMethods
+        MODULE_NAME = :DynamicMoneyAttributes
+
         def monetized_attributes
           monetized_attributes = @monetized_attributes || {}.with_indifferent_access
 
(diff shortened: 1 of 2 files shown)


sunny (dismissed):
Neat! Can you add an entry to the CHANGELOG as well?

sunny (dismissed):
Tests don’t seem to run though, but it doesn’t seem to be linked to your changes 🤔

jbennett:
Looks like #698 also had issues with sqlite3 and added a `~> 1.4` requirement to the main gemfile. Should I look to do the same?

sunny (approved):
❤️

sunny (approved):
It's been ages, but could you please resolve the conflicts since? 🙏🏻

PR #706, opened 2024-10-11
Monetizable: Ignore converted methods when writing attributes

edithemmings:
Discovered an issue where making an update to instance currencies causes a MissingAttributeError for converted methods

Example model:
```ruby
class Transaction < ActiveRecord::Base
  monetize :amount_cents, with_model_currency: :currency
  monetize :tax_cents, with_model_currency: :currency
  monetize :total_cents, with_model_currency: :currency
  monetize :optional_amount_cents, with_model_currency: :currency, allow_nil: true

  def total_cents(foo = 0, bar: 0)
    amount_cents + tax_cents + foo + bar
  end
end
```

Triggering the error:
```
transaction = Transaction.create(currency: 'USD')
transaction.update(currency: 'CAD')
ActiveModel::MissingAttributeError: can't write unknown attribute `total_cents`
```

It traces to line 208. 

I updated all calls to `write_attribute` in Monetizeable to check that the monetized attribute is a true attribute before trying to write it.


sunny (approved):
Awesome! Could you add a small entry to the Changelog as well? 🙏🏻

sunny (approved):

yukideluxe (approved):
This is great! Thanks for the fix!

PR #712, opened 2025-01-24
Fix thousand separator issue in validation

RubenIgnacio:
* Organize monetizable specs to improve readability.
* Add tests to validate a monetizable field that uses a currency with a dot as a thousand separator.
* Fix an issue when validating a monetizable field that uses a currency with a dot as a thousand separator.


spec/active_record/monetizable/shared_contexts.rb
```diff
@@ -0,0 +1,16 @@
+RSpec.shared_context "monetizable product setup" do
```
sunny: Could we simply copy this product in the two locations it is used? I find that specs are easier to understand when their context is right under your nose, even if it leads to duplication.

spec/active_record/monetizable/money_helpers.rb
```diff
@@ -0,0 +1,91 @@
+module MoneyHelpers
```
sunny: My two cents: I prefer longer specs to the extra indirection of adding these custom expectation methods. But I understand that this is quite a personal preference.

sunny: Yeah I'd tend towards removing them, thank you! 🙏🏻

RubenIgnacio: I've already remove the `MoneyHelper`

RubenIgnacio:
I've been reviewing my fix and realized that the issue was specific to my local setup. There are existing configurations that solve this without needing these changes, so I'll close this PR. Sorry for the trouble!

I also noticed that those methods could be moved to the `Details` class to improve clarity.

image

PR #705, opened 2024-08-30
Add support for Rails 7.1 and 7.2

schmijos:
This adds two new Gemfiles and bumps minimal versions.

PR #716, opened 2025-07-09
Fix CI

yukideluxe:
Hello everybody!

The first action one must take when it comes to maintenance is ensuring that CI is green so here we are 😬

This PR does the following:

- Officially drops support for Rails < 6.1 and Ruby < 3.0. We haven’t been testing against Ruby 2.7 since https://github.com/RubyMoney/money-rails/pull/692. To me, that means we stopped supporting it in May 2024, so there's no point in testing against Rails 5 since [it's not compatible with Ruby >= 3(https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html).
- Adds tests against Ruby 3.4 💪🏻 
- Includes an easy bump of the test gems 😊

@sunny I’d really appreciate your input here! 😊 cc @tjschuck

PS: Fixing MongoDB tests comes next. They’re currently not running because there’s a check that skips them if Ruby < 3. ✌🏻


Rakefile
```diff
@@ -28,22 +26,29 @@ task spec: :prepare_test_env
...
+  invoke_db_task("create")
+  invoke_db_task("migrate")
+  invoke_db_task("test:prepare")
 end
 
 def run_with_gemfile(gemfile)
   Bundler.with_original_env do
-    begin
-      sh "BUNDLE_GEMFILE='#{gemfile}' bundle install --quiet"
-      Rake.application['app:db:create'].invoke
```
yukideluxe, line 41: before running the tests, the `prepare_test_env` is run so I don't think we need to do this here 🤔

Rakefile
```diff
@@ -28,22 +26,29 @@ task spec: :prepare_test_env
...
-      Rake.application['app:db:test:prepare'].invoke
-      sh "BUNDLE_GEMFILE='#{gemfile}' bundle exec rake spec"
-    ensure
-      Rake.application['app:db:drop:all'].execute
-    end
+    lockfile = "#{gemfile}.lock"
+    File.delete(lockfile) if File.exist?(lockfile)
```
yukideluxe, line 39: making sure gems are properly updated in case the specific gemfile changes 😊

spec/configuration_spec.rb
```diff
@@ -138,7 +138,7 @@
 
       context "when passed an invalid value" do
         it "should raise an ArgumentError" do
-          expect(lambda{MoneyRails.rounding_mode = "booyakasha"}).to raise_error(ArgumentError, 'booyakasha is not a valid rounding mode')
+          expect { MoneyRails.rounding_mode = "booyakasha" }.to raise_error(ArgumentError, 'booyakasha is not a valid rounding mode')
```
yukideluxe, line 141: I noticed this was also fixed in this PR! https://github.com/RubyMoney/money-rails/pull/705. It's just a deprecation warning 😊

gemfiles/rails6.1.gemfile
```diff
@@ -2,6 +2,10 @@ source 'https://rubygems.org'
 
 gem 'rails', '~> 6.1.0'
 gem 'i18n',  '~> 1.6'
+gem "bigdecimal"
```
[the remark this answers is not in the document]
sunny: If we wanted to simplify this a bit we could add all the common dependencies to the gemfile as dev dependencies, but they might not make sense for the mongoid gemspecs.

I’d say keeping the small duplication here would be fine as well.

spec/dummy/config/boot.rb
```diff
@@ -1,13 +1,15 @@
 require 'rubygems'
 
-# The default gemfile is rails4
-gemfile = File.expand_path('../../../../gemfiles/rails4.gemfile', __FILE__)
+# Default to rails7.0.gemfile if BUNDLE_GEMFILE is not set
+default_gemfile = File.expand_path('../../../../gemfiles/rails7.0.gemfile', __FILE__)
```
sunny, line 4: Minor simplification suggestion while we’re at it.

```suggestion
default_gemfile = File.expand_path('../../../gemfiles/rails7.0.gemfile', __dir__)
```

sunny (approved):
Yay! ✨

yukideluxe:
Thanks for the review @sunny, I applied your suggestions! We are still figuring out the repo permissions, but @sunny approve counts as an unblocking approve to me, so I am merging so I can unblock some open PRs 😬

PR #717, opened 2025-07-10
Add minimum required Ruby version

tagliala:
Ruby 2.5 has been an implicit requirement due to activesupport/railties 6.1, and this commit makes it explicit in the gemspec for consistency with Money, clarity, and accurate version display on https://rubygems.org/gems/money-rails, which now shows `>= 0`.

The encoding magic comment is therefore removed, as Ruby 2 defaults to UTF-8 encoding.

Autofixed with:
```
rubocop -a --only Style/Encoding
```


yukideluxe:
tl;dr I think we should set the minimum Ruby requirement against >= 3.1 because the older versions are past their EOL

@sunny and @tagliala, I had a conversation with @tjschuck yesterday about dropping support for Ruby versions that are past their EOL, and our philosophy is "if it can still work, let’s keep it working, because then everyone can benefit" but as soon as we cannot move forward with improvements "fuck it" 😂 

We are [not testing with Ruby < 3.0](https://github.com/RubyMoney/money-rails/pull/716), so, as of today, the minimum requirement should be Ruby 3 🙏🏻 

Apart from that, these test improvements https://github.com/RubyMoney/money-rails/pull/705 (ping @schmijos) are [not passing against Ruby 3.0](https://github.com/RubyMoney/money-rails/actions/runs/16216895552/job/45789663125?pr=705), so I am very much inclined to drop support for 3.0 completely because it will make it complicated for us to keep on doing improvements and complicate the codebase 🙆🏻‍♀️ 

What do you think!?

yukideluxe (approved):

PR #720, opened 2025-07-12
Improve job name for clarity in workflow runs

tagliala:
# ⚠️ CI-settings breaking change. Settings should be changed to require status to pass with the new name

-----

Previously, the job name in the workflow matrix only displayed version numbers, resulting in ambiguous labels like

```
Ruby / test (3.1, 4.4, 3.0)
```

This commit updates the job name to explicitly include the technology names alongside their versions:

```
CI / Test (Ruby 3.1, MongoDB 4.4, SQLite 3.0)
```

This change makes it immediately clear which version number corresponds to which technology, improving readability and reducing confusion when reviewing workflow results.

---

image


yukideluxe (approved):
thanks! I'll tweak the settings no problemo!

PR #726, opened 2025-12-13
Remove legacy code for old Rails and Ruby versions

c960657:
Clean up legacy support code.

- Bump Rails version requirement to 7.0.
- Remove support code for old versions of Rails, Ruby and Mongoid.
  - Rails 7 requires Mongoid >= 7.3.4, so remove mongoid6.gemfile.
  - `Hash#deep_symbolize_keys` was introduced in Rails 4.
- Add test gemfiles for Rails 8.0 and 8.1.
- Enable JRuby on GitHub Action (apparently activerecord-jdbc-adapter is not yet compatiable with Rails 8, so it skips those tests).
- Enable Database Cleaner for Mongoid.
- Adjust Rakefile so tests don't run twice.

Rails 7.0 and Rails 7.1 are both EOL, but there is no additional effort in supporting those, so I kept those for now.

Ruby 3.1 is also EOL. There is just a single line in Rakefile to support this.

These requirements are very conservative and in line with the money gem. They can of course easily be bumped even more, but I suggest they stay in sync with the money gem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased
 - **Breaking change**: Drop support for Rails < 7.0
 
+- Fix `MoneyRails::ActionViewExtension` not available during eager loading (GH-614)
+
 ## 2.0.0
 
 - **Breaking change**: Requires `money` gem version ~> 7.0.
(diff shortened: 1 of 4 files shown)


c960657:
Oh, hang on, I forgot to fix the last test in spec/mongoid/mongoid_spec.rb.

yukideluxe:
I loooove this! 🤩 we'll get back to you soon!

PR #727, opened 2025-12-15
Fix ActionViewExtension not available during eager loading

### LINKED ISSUE #614: uninitialized constant MoneyRails::ActionViewExtension after upgrading to Rails 6.1
I call "include MoneyRails::ActionViewExtension" in a couple helpers and Prawn pdf documents so I can use humanized_money_with_symbol and other money methods.  This has worked as expected for years...until today.

Upon upgrading to Rails 6.1, I now get the following error:

      NameError - uninitialized constant MoneyRails::ActionViewExtension

I'm assuming this is the result of a structural change in Rails 6.1.  I am still figuring out this change, and others....but, thought this was worthy of submitting.

Sorry I don't have more to offer by way of details or solutions...but, I'm not sure how these pieces all work together.

fcatuhe:
## Summary

- Fix `MoneyRails::ActionViewExtension` not being available during eager loading (`config.eager_load = true`)
- Use Ruby's `autoload` to define the constant immediately at require time, deferring file loading until first access

## Problem

When a Rails app has `config.eager_load = true` (production), referencing `MoneyRails::ActionViewExtension` raises:

```
uninitialized constant MoneyRails::ActionViewExtension
```

This happens because the extension is only loaded inside an `ActiveSupport.on_load(:action_view)` callback. During eager loading, app files (like decorators) that `include MoneyRails::ActionViewExtension` are loaded before ActionView triggers the callback.

## Solution

Use `autoload` to register the constant immediately while deferring the actual file load:

```ruby
module MoneyRails
  autoload :ActionViewExtension, 'money-rails/helpers/action_view_extension'
end
```

The `on_load(:action_view)` callback still includes the helper into `ActionView::Base`, but no longer needs to handle the require.

## Test plan

- Verified existing test suite passes
- Can be tested by creating a decorator that includes `MoneyRails::ActionViewExtension` in an app with `config.eager_load = true`

Fixes #614

### DIFF
--- a/spec/helpers/action_view_extension_spec.rb
+++ b/spec/helpers/action_view_extension_spec.rb
@@ -1,5 +1,13 @@
 require 'spec_helper'
 
+describe 'MoneyRails::ActionViewExtension' do
+  describe 'autoload' do
+    it 'is registered via autoload for eager loading compatibility' do
+      expect(MoneyRails.autoload?(:ActionViewExtension)).to eq('money-rails/helpers/action_view_extension')
+    end
+  end
+end
+
 describe 'MoneyRails::ActionViewExtension', type: :helper do
   describe '#currency_symbol' do
     subject { helper.currency_symbol }


sunny (commented):
Awesome fix. Can you add a CHANGELOG entry as well? 🙏🏻

fcatuhe:
Done and ready for review.

PR #772, opened 2026-03-29
Remove unneeded check on `ActiveModel#attribute_aliases`

wwahammy:
As far as I can tell, `ActiveModel#attribute_aliases` has been in Rails since at least version 6 so there shouldn't be a reason to check whether the method is on the record class. The removed comment implies it was added in Rails 4 but who knows? In regards to the current supported versions: here's the [file](https://github.com/rails/rails/blob/f4321f78fe567fd29dd0a0fb5cb17af68efd36bb/activemodel/lib/active_model/attribute_methods.rb) included in on Rails 7 and here's the [file](https://github.com/rails/rails/blob/64b90c735614681fcf6e2e7f0022c8a9d2f8554f/activemodel/lib/active_model/attribute_methods.rb) for Rails 8.

This removes the unnecessary check.

### DIFF
--- a/lib/money-rails/active_record/monetizable.rb
+++ b/lib/money-rails/active_record/monetizable.rb
@@ -271,12 +271,9 @@ module MoneyRails
           # We haven't defined our own subunit writer, so we can invoke
           # the regular writer, which works with store_accessors
           public_send("#{subunit_name}=", money.try(:cents))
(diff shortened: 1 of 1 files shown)


sunny (approved):

yukideluxe:
thanks!

yukideluxe (approved):

PR #780, opened 2026-05-27
Fix MoneySerializer not registered on Rails 8.1.0+

### LINKED ISSUE #779: MoneySerializer not loaded in Rails 8.0+
In Rails >= 8.0, the hook in `lib/money-rails/hooks.rb` to load the serializer is wrong, it's using `:active_job` and should use `:active_job_arguments`

yukideluxe:
Fixes https://github.com/RubyMoney/money-rails/issues/779

Hello @sunny and whoever is watching 👋🏻 Claude and I had a stab at the reported issue and this is what we've come up with!  

Seems that on Rails 8.1+, `MoneyRails::ActiveJob::MoneySerializer` is not registered with Active Job, so serializing a `Money` job argument fails.

### Root cause

money-rails registered its serializer by appending to `config.active_job.custom_serializers` *inside* an
`ActiveSupport.on_load(:active_job)` block. Rails reads that list at a version-dependent moment:

| Rails | Consumes `custom_serializers` via |
|-------|-----------------------------------|
| ≤ 8.0 | `config.after_initialize` |
| 8.1.0 / 8.1.1 ([rails/rails@b7cd3018ef](https://github.com/rails/rails/commit/b7cd3018ef)) | `on_load(:active_job)` |
| 8.1.2+ ([rails/rails@c0dc92e9e9](https://github.com/rails/rails/commit/c0dc92e9e9)) | `on_load(:active_job_arguments)` |

Rails <= 8.0 read `config.active_job.custom_serializers` in `after_initialize` (end of boot), so money-rails's appended serializer was always there in time.

8.1.0 ([b7cd3018ef](https://github.com/rails/rails/commit/b7cd3018ef)) moved that read into `on_load(:active_job)` — the **same hook money-rails uses**. Load-hook callbacks run in registration order, and Rails registers before money-rails (a gem). So Rails reads the list while it's still empty, then money-rails appends too late .

8.1.2 changed Rails' read to `on_load(:active_job_arguments)` but changing money-rails to use that hook doesn't help. First, it is not compatible with older Rails version and, second, the hook could potentially be called before the custom serializer is in the list.) would break on 7.0–8.1.1.

### Fix

Register the serializer during initialization instead of inside a load hook, so it is present `config.active_job.custom_serializers` before Rails reads the list regardless of how Rails consumes it.

I added a simple integration test that round-trips a `Money` through `ActiveJob::Arguments`

@jaredmoody would you mind checking if this fixes what you are seeing? thanks!

### DIFF
--- a/lib/money-rails/hooks.rb
+++ b/lib/money-rails/hooks.rb
@@ -43,14 +43,20 @@ module MoneyRails
         ::ActionView::Base.include MoneyRails::ActionViewExtension
       end
 
-      # For ActiveSupport
-      ActiveSupport.on_load(:active_job) do
-        if defined?(::ActiveJob::Serializers)
-          require "money-rails/active_job/money_serializer"
-          Rails.application.config.active_job.tap do |config|
-            config.custom_serializers ||= []
-            config.custom_serializers << MoneyRails::ActiveJob::MoneySerializer
-          end
+      # For Active Job
+      #
+      # Registered during initialization rather than inside an
+      # `ActiveSupport.on_load(:active_job)` block so the serializer is present in
+      # `config.active_job.custom_serializers` before Rails consumes the list. Rails
+      # reads it at different points across versions (`after_initialize` on <= 8.0,
+      # `on_load(:active_job)` on 8.1.0/8.1.1, `on_load(:active_job_arguments)` on 8.1.2+),
+      # any of which an `:active_job` hook here could lose the race to. See issue #779 and
+      # rails/rails commits b7cd3018ef (8.1.0) and c0dc92e9e9 (8.1.2).
+      if defined?(::ActiveJob::Serializers)
+        require "money-rails/active_job/money_serializer"
+        Rails.application.config.active_job.tap do |config|
+          config.custom_serializers ||= []
+          config.custom_serializers << MoneyRails::ActiveJob::MoneySerializer
         end
       end
     end
--- a/lib/money-rails/hooks.rb
+++ b/lib/money-rails/hooks.rb
@@ -44,20 +44,9 @@ module MoneyRails
       end
 
       # For Active Job
-      #
-      # Registered during initialization rather than inside an
-      # `ActiveSupport.on_load(:active_job)` block so the serializer is present in
-      # `config.active_job.custom_serializers` before Rails consumes the list. Rails
-      # reads it at different points across versions (`after_initialize` on <= 8.0,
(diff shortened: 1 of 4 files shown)


lib/money-rails/hooks.rb
```diff
@@ -43,14 +43,20 @@ def self.init
...
+      # For Active Job
+      #
+      # Registered during initialization rather than inside an
+      # `ActiveSupport.on_load(:active_job)` block so the serializer is present in
+      # `config.active_job.custom_serializers` before Rails consumes the list. Rails
+      # reads it at different points across versions (`after_initialize` on <= 8.0,
+      # `on_load(:active_job)` on 8.1.0/8.1.1, `on_load(:active_job_arguments)` on 8.1.2+),
+      # any of which an `:active_job` hook here could lose the race to. See issue #779 and
+      # rails/rails commits b7cd3018ef (8.1.0) and c0dc92e9e9 (8.1.2).
+      if defined?(::ActiveJob::Serializers)
```
wwahammy, line 47: Is it possible for ActiveJob::Serializers to be defined and `Rails.application.config.active_job` to not be available? I don't know how but I wonder.

sunny (dismissed):

lib/money-rails/hooks.rb
yukideluxe: thanks for raising this! not really sure but I am proposing another solution now! Let me know!

yukideluxe:
@jaredmoody [reported](https://github.com/RubyMoney/money-rails/issues/779#issuecomment-4846860880) my original fix did not solve the problem and @sk- [said](https://github.com/RubyMoney/money-rails/issues/779#issuecomment-4845745596) the current code doesn't work when `config.eager_mode = false`, so I've been drilling Claude to test all the combos 😳 

I must admit this is going a bit over my head and my main goal is to simplify the code and **avoid Rails version checks at all costs**, so after a bunch of back and forth I think the whole thing can be simplified by just adding the `MoneySerializer` directly into `::ActiveJob::Serializers` and stop using the `custom_serializers` config that is consumed at different moments of boot time depending on the Rails version used... I think this also addresses @wwahammy's [concern](https://github.com/RubyMoney/money-rails/pull/780#discussion_r3312954548) 🤔


Could somebody help me confirm?! 🙏🏻

Ping @sunny

PS: CI failures seem unrelated, I will fix in `main` before merging

wwahammy (commented):
@yukideluxe I think this all makes sense to me. I think it looks good.

One concern I do have is that the the CI only appeared to test on MongoDB so I'd be worried about whether the tests are missing anything due to that.

Issue #291, opened 2015-02-18
How do I load money-rails properly outside of rails?


konung:
Hello.

### Background: 

I'm trying to use Money-rails out side of rails. Specifically I'm using praxis framework ( it's ruby based api framework - https://github.com/rightscale/praxis ) . Obviously it's not rails  - However I'm using ActiveRecord 4.2 as my ORM and ActiveSupport as well 4.2 . My DB is MySQL 5.5 . I'm able to use other ActiveRecord related gems like ransack, caminari & standalone_migrations

### Problem
Even thou money-rails gets loaded in my environment - when I try to do: 

````ruby
class SalesOrderLine < ActiveRecord::Base
   .....
   monetize :price_cents

end
````

I get:
````ruby
NoMethodError: undefined method `monetize' for #<Class:0x007f9e4f731d48>
````
Any suggestions?

### Temporary solution

I sidestepped the issue for now like so, so I know Money and Monetize gems are loaded and working properly

````ruby
  def price
    Money.new(self.price_cents)
  end

  def price=amount
    self.price_cents = Monetize.parse(amount).cents
  end
````

**Comment by semmons99** [comment]
My guess, off-hand, is there's some autoloading happening in rails and not praxis. I think you just need to do this: https://github.com/RubyMoney/money-rails/blob/master/lib/money-rails.rb#L13

**Comment by stevenbarragan** [comment]
Same problem here, @konung did you find a solution?

**Comment by stevenbarragan** [comment]
I was having problems to test my gem (where I'm using ``money-rails``)

I put this into an initializer and it works, I hope it helps

```ruby
MoneyRails::Hooks.init
```

semmons99:
Closed due to inactivity.

PR #786, opened 2026-09-02
Register framework integrations when the gem is required

mokevnin:
## What

Move the two framework mixins — `MoneyRails::ActiveRecord::Monetizable` and `MoneyRails::ActionViewExtension` — from `MoneyRails::Hooks.init` to require time in `lib/money-rails.rb`.

## Why

`Hooks.init` is only called from the railtie initializer, and it is what registers the `ActiveSupport.on_load(:active_record)` / `on_load(:action_view)` blocks. Outside a full Rails boot nothing registers them, so `monetize` ends up defined on no class at all:

```ruby
require "active_record"
require "money-rails"
ActiveRecord::Base.respond_to?(:monetize) # => false
```

That makes the Active Record integration invisible to everything that inspects the code rather than boots a server: doc generators, editor autocomplete, static analysis, and a plain `require "money-rails"` in a script or a test harness. It is the same class of problem as #614 / #727, where `MoneyRails::ActionViewExtension` was unavailable until a callback happened to fire.

Registering an `on_load` hook is lazy and cheap — it does not load Active Record or Action View, it only queues the block until something else loads them. So people using the gem without Rails (#470, #291) pay nothing for the registration, and get `monetize` from the bare `require`.

`require "active_support/lazy_load_hooks"` is added explicitly rather than relying on `money`/`monetize` to have pulled Active Support in; `activesupport` is already a runtime dependency.

## What stays put

The migration extensions stay in `Hooks.init`. Choosing between the PostgreSQL and generic variants reads `connection_db_config`, which needs a configured database — exactly what the railtie's `after: "active_record.initialize_database"` guarantees, and not something a bare `require` can promise.

`Hooks.init` remains public and its README-documented use for non-Rails apps is unchanged. The mixins it no longer performs are already registered by the `require` it necessarily follows, and a repeated `include` of the same module is a no-op regardless.

No behavior change — Rails apps load the same modules into the same classes, in the same order.

## Tested

- `bundle exec rake spec` green on `active_record7.0`, `7.1`, `7.2`, `8.0` and `8.1` (Ruby 4.0), 278 examples each.
- `bundle exec rubocop` clean.
- The new spec shells out to a fresh Ruby that requires `money-rails` *before* `active_record`, asserts the require did not eagerly load Active Record, then asserts `monetize` is there. It fails on `main` and passes here.
- Could not run the Mongoid gemfiles (no local mongod) or the JRuby cell.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Register the Active Record and Action View integrations when the gem is required, instead of from the railtie initializer
+
 ## 3.0.0
 
 - **Breaking change**: Drop support for Rails < 7.0
--- a/lib/money-rails.rb
+++ b/lib/money-rails.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "active_support/lazy_load_hooks"
 require "money"
 require "monetize"
 require "monetize/core_extensions"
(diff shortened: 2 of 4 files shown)


sunny (approved):
Makes sense!

mokevnin:
Rebased onto `main` so the branch is no longer behind — it was only #787 (binstubs), zero overlap with this diff, so the change itself is byte-identical to what you approved.

Verified locally on the rebased commit: `bin/rake spec` (278 examples, 0 failures) and `bin/rubocop` (77 files, no offenses).

CI is sitting at `action_required` — it needs a maintainer to approve the workflow run.

PR #788, opened 2026-09-13
Respect top-level null option in migration helpers

### LINKED ISSUE #775: null option in migration helper is (still) ignored
in a migration:

```ruby
create_table ...
  t.monetize :price, null: true
end
```

this will not lead to a database field that allows (it is still `NOT NULL`). This is the same, buggy, behavior as described in #189 

Current workaround is to specify the migration with an explicit integer field (for the `_cents` field). 

Tested with

money-rails 3.0.0
money 7.0.2
rails 8.1.3

cpruijsen:
`add_monetize :products, :price, null: true` silently ignores the `null` option. Only the per-column
forms, `amount: { null: true }` and `currency: { null: true }`, have any effect, so a migration
written the shorter way produces NOT NULL columns with no indication that the option was dropped.

`OptionsExtractor.extract` in
`lib/money-rails/active_record/migration_extensions/options_extractor.rb` merged the configured column
defaults with `options[attribute]` and never looked at the top level. It now merges a top-level `null`
in between the two, so it applies to both columns while a per-column option still overrides it. The
value is compacted away when absent, so a migration that passes no `null` behaves exactly as before.

README documents the top-level form alongside the per-column one, and says which wins, since the
precedence is the part that is not obvious.

Fixes #775


lib/money-rails/active_record/migration_extensions/options_extractor.rb
```diff
@@ -5,7 +5,11 @@ module ActiveRecord
     module MigrationExtensions
       class OptionsExtractor
         def self.extract(attribute, table_name, accessor, options = {})
-          default = MoneyRails::Configuration.send("#{attribute}_column").merge(options[attribute] || {})
+          # A top-level `null` option applies to both columns; per-attribute
+          # `amount`/`currency` options take precedence over it.
+          default_for_both_columns = { null: options[:null] }.compact
+          default = MoneyRails::Configuration.send("#{attribute}_column")
+                                             .merge(default_for_both_columns, options[attribute] || {})
```
sunny, line 12: Minor syntax preference here to avoid having lines that are too long:
```suggestion
          default = default.merge(default_for_both_columns, options[attribute] || {})
```

sunny (commented):
Good call.

Could you add an entry to the CHANGELOG as well please? 🙏🏻

cpruijsen:
Added in the latest commit, under Unreleased. Thanks for the review.

lib/money-rails/active_record/migration_extensions/options_extractor.rb
yukideluxe: agree with this but nothing we cannot fix with a rubocop rule 😊

yukideluxe:
Hello, @cpruijsen! First, thanks for the PR! 🎉 

Second, I disagree with the `null` option is dropped with "no indication". It might not be what you expect if you're used to the standard Rails migration helpers, but it's just how the monetize helper works. It seems that it doesn't read top-level options, by design. You can already do `amount: { null: true }, currency: { null: true }` to achieve exactly the same result today, unless I am missing something 🤔

There could also be confusion around the other Rails migration options that are equally ignored at the top level but can be passed per-column, like `limit`, `default` and `precision`. This PR would allow `null` at the top level but not the others, and I am not sure whether allowing some is better than allowing none, since most of them can't be shared anyway because the two columns have different types. For example, we could allow `default: nil`, but no other default value would make sense.

**Maybe this is just a documentation problem?**

@sunny, what do you think?

PS: sorry, I've been a bit MIA, but I'm baaaack

sunny:
While we could add some documentation to explain that you can do `amount: { null: true }, currency: { null: true }`, I think calling `null: true` makes more sense for both at the same time and would rather encourage that by giving it a shorter syntax as it is common to want to nullify a price column.

cpruijsen:
Fair on the framing, "no indication" overstates it. `amount: { null: true }, currency: { null: true }` does work today, and I should have said so in the description.

On the substance this is what @sunny describes: top-level `null` as a shorter way to say the same thing, with per-attribute options still taking precedence, so nothing existing changes.

On your point about the other top-level options, I kept this to `null` deliberately. Passing all of them through would change behaviour for anyone relying on the current merge, and #775 only asked for `null`.

Happy either way. If you would rather document the per-attribute form, say so and I will close this and send a README patch instead.

yukideluxe:
> While we could add some documentation to explain that you can do `amount: { null: true }, currency: { null: true }`, I think calling `null: true` makes more sense for both at the same time and would rather encourage that by giving it a shorter syntax as it is common to want to nullify a price column.

okok, I'm not against this change, it's just a shortcut to something that it's already short IMHO :P

The documentation is already there (and it has been extended in this PR!), it could be a bit clearer but it can be tweaked after this is merged ;)

@cpruijsen could you tweak the CHANGELOG to say we are just adding this option, though? 🙏🏻

